PR #68, opened 2016-07-04
Fixes #66 Not selectable objects depending on the order of components

### LINKED ISSUE #66: Loaded models (dae or obj) are not selectable in 3D view
Works fine if you use the scene graph to select.

fernandojsg:


fernandojsg:
@caseyyee can you git a try with your issue https://github.com/aframevr/aframe-editor/issues/66#issuecomment-230347104 ?

caseyyee:
@fernandojsg doesn't seem to work for me still.  

Here is the markup i'm using:

``` html
    <a-scene>
      <a-assets>
        <a-asset-item id="chicken" src="models/chicken/ayam.dae"></a-asset-item>
      </a-assets>

      <a-entity collada-model="#chicken" position="0 1 0"></a-entity>

      <a-plane rotation="-90 0 0" width="4" height="4" color="#7BC8A4"></a-plane>

      <a-sky color="#ECECEC"></a-sky>
    </a-scene>

```

The issue I filed was specific to using models only.  I haven't tested for the order of components.

fernandojsg:
I fixed it by including a listener to  `model-loaded`.

PR #88, opened 2016-07-05
Scenegraph tweaks

caseyyee:
image
1. Add entity now in scene graph header (more obvious place)
2. Removes extraneous bullets, but leaves useful icons for key scene entity types (camera, lights)
3. Outliner now scrolls.  Bottom bar remains fixed.

Resolves issues: 
https://github.com/aframevr/aframe-editor/issues/87
https://github.com/aframevr/aframe-editor/issues/37
https://github.com/aframevr/aframe-editor/issues/6
https://github.com/aframevr/aframe-editor/issues/15
 https://github.com/aframevr/aframe-editor/issues/16

### DIFF
--- a/src/css/dark.css
+++ b/src/css/dark.css
@@ -89,27 +89,27 @@ input.number:focus {
 	border-bottom: 1px solid #111;
 }
 
-	#menubar .menu {
-		float: left;
+	.menu {
+		display: inline-block;
 		cursor: pointer;
 		padding-right: 8px;
 	}
 
-	#menubar .menu.right {
+	.menu.right {
 		float: right;
 		cursor: auto;
 		padding-right: 0;
 		text-align: right;
 	}
 
-		#menubar .menu .title {
+		.menu .title {
 			display: inline-block;
 			color: #888;
 			margin: 0;
 			padding: 9px;
 		}
 
-		#menubar .menu .options {
+		.menu .options {
 			z-index: 1500;
 			position: absolute;
 			display: none;
@@ -118,31 +118,31 @@ input.number:focus {
 			width: 150px;
 		}
 
-		#menubar .menu:hover .options {
+		.menu:hover .options {
 			display: block;
 		}
 
-			#menubar .menu .options hr {
+			.menu .options hr {
 				border-color: #222;
 			}
 
-			#menubar .menu .options .option {
+			.menu .options .option {
 				color: #888;
 				background-color: transparent;
 				padding: 5px 10px;
 				margin: 0 !important;
 			}
 
-				#menubar .menu .options .option:hover {
+				.menu .options .option:hover {
 					color: #fff;
 					background-color: #08f;
 				}
 
-				#menubar .menu .options .option:active {
+				.menu .options .option:active {
 					background: transparent;
 				}
 
-		#menubar .menu .options .inactive {
+		.menu .options .inactive {
 			color: #444;
 			background-color: transparent;
 			padding: 5px 10px;
@@ -155,7 +155,6 @@ input.number:focus {
 	top: 33px;
 	bottom: 0;
 	width: 180px;
-	overflow: auto;
 	background: #2b2b2b;
 }
 
@@ -413,14 +412,14 @@ div.vec3 {
 	left: 0;
 }
 
-a.button {
+.button {
 	color: #bcbcbc;
 	font-size: 16px;
 	text-decoration: none;
 	margin-left: 10px;
 }
 
-a.button:hover {
+.button:hover {
 	color: #1faaf2;
 }
 
--- a/src/css/dark.css
+++ b/src/css/dark.css
@@ -3,11 +3,14 @@
 	background: #2b2b2b;
 	padding: 0;
 	width: 100%;
-	font-size: 12px;
+  margin: 0;
+  font-size: 12px;
 	cursor: default;
-	overflow: auto;
 	outline: none;
 	border-top: 1px solid #111;
+  flex: 1 1 auto;
+  overflow-y: auto;
+  min-height: 0px;
 }
 
 	.Outliner .option {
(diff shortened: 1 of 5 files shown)


caseyyee:
@fernandojsg ahg, overlapping work already (lol, my first go at changes -- go figure) https://github.com/aframevr/aframe-editor/pull/86

src/css/dark.css
```diff
@@ -89,27 +89,27 @@ input.number:focus {
 	border-bottom: 1px solid #111;
 }
 
-	#menubar .menu {
-		float: left;
+	.menu {
+    display: inline-block;
```
caseyyee, line 115: prob need to switch the whitespace in this doc.  will do in another commit.

dmarcos:
Let's keep the icons as stated in the design document. The icons are useful to give an overview of what components the entity contains.

dmarcos:
PRs should be focused and target a single issue or changes are very very hard to follow and review

dmarcos:
We can revisit design decisions in the future but now we should focus on polishing the current design instead of debating the fundamentals. The design document has been circulating for two months and now there's no time to discuss it.

ngokevin:
If the PR becomes focused to just removing the icons, I agree with that. The icons aren't very useful given that they only work for camera/geometry/material/light, and it just creates noise.

We can asynchronously discuss it in this thread if we disagree, but it's counter-productive to form a wall around the editor.

dmarcos:
They don't create noise they give you a quick overview of the entity. It's a common patter in popular creation tools like cinema 4D or 3D Studio Max:
Cinema 4D
3D Studio Max

dmarcos:
Let's focus on polish and ship the editor. It's out of the current scope to discuss this kind of design decisions. As I said, the spec document have been circulating for months now and nobody made any comment.

caseyyee:
The icons are not useful and are just confusing.   How is anyone supposed to know what a painting is supposed to be?  Or a box?   It's too ambiguous.   c4d or max is not the right comparison and shouldn't be taken as a reference point.  those tools are for a different purpose and audience.

ngokevin:
This can be considered part of polish. It seems very arbitrary to call it "out of scope" or a "design decision". Given that it is a small change that has valid reasons and at least a couple of people agree, I don't see the issue. Hopefully, this does not stem from past disagreements.

Surprisingly, you can miss minor details in a large document that only arise once you actually try it. We shouldn't ever consider any plans to be set in stone. For example, check the 0.3.0 planning whiteboard in the SF office; it says _Universal Controls_. 

Just because there's a reason does not mean it's a good reason; that's "programmer design". Cinema4D and 3ds Max are not ECS-based Editors, they are modeling programs where each object has a distinct type they can easily represent a single icon for each object. For A-Frame, an entity can have infinite components, and right now all we are showing is a picture of a box and a picture frame, which is not nearly as descriptive as the ID. This encourages developers to name their entities for more readable code.

dmarcos:
Those icons are learned from experimentation and become familiar once you start using the tool.

dmarcos:
Universal Controls was never part of the scope of 0.3.0 and I mentioned that multiple times.

dmarcos:
The way to navigate, manipulate and explore a scenegraph in the a-frame editor, cinema4D and 3D Max is very similar even it the overall goal of the tool is different

ngokevin:
- That's a false assumption. The icons are not useful theoretically, and in practice as we **are using** the tool.
- There are two places where 0.3.0 is explicitly documented. Check what is on them both:
  - https://github.com/aframevr/aframe/milestone/2
  - The SF whiteboard
- Navigating/manipulating a scenegraph is separate from showing non-helpful icons.

dmarcos:
The icons help you understand and navigate the scene. You could click on the camera or material icon and direclty expand an highlight the corresponding component. It helps discriminate visually the entities that impact the rendered scene vs those that are just structural or for layout purposes. It's not guaranteed that all the entities are going to have IDs or a mixin. You would often see just a tree of `a-entity` It can also give you info about the component by just hovering on it without having to click or go to the component panel.

dmarcos:
@ngokevin @caseyyee The purpose of a spec document is to explore different design solutions in a cheap manner without having to pay the implementation costs. It's true that there's new information you discover once you start using the tool but in this case I feel the conversation about icons could have easily taken place months ago.

dmarcos:
There are ton of things than need polish and improvement before release. Any decision can be questioned for the next version but for now let's stick with what it's speced and postpone this conversation.

dmarcos:
As a general comment. A UI feature that is not immediately obvious for a novice it doesn't have to be synonym of bad usability or clutter. Good tools are learnable and while accessible to novices they also provide convenience flows for more expert users. The scenegraph icons fall in this bucket. As a complet novice you won't probably understand or use them but overtime they are learned and become a more efficient way to do certain operations. I recommend reading all the [research on marking menus](http://www.billbuxton.com/MMUserLearn.html) and how good UIs grow and accommodate to the level of expertise of the user

fernandojsg:
Regarding the icons discussion I've answered in the issue itself, I believe it will be better to have it there so everyone can follow the opinions.
I believe we should keep doing one PR per issue so we don't block some features while discussing others like this case.
- I agree with removing the dots (https://github.com/aframevr/aframe-editor/issues/6) as they don't add any value and we could use that spot for collapsing arrows for example.
- The sticky header is great to have yep (https://github.com/aframevr/aframe-editor/issues/16)
- Regarding moving the `Create to scenegraph` (https://github.com/aframevr/aframe-editor/issues/15) in my PR (https://github.com/aframevr/aframe-editor/pull/86, and yep I need to split that one too) I've added the plus sign on the bottom as we'll have more buttons and it will be better to have a bar to include all them, actually when I first saw the screenshot in this PR without reading the text, I thought that you included that `[+]` to do `collapse/expand All nodes`. I don't really like the placement there, as I said we'll have more icons so better to keep then together somewhere like in the status bar.

Regarding the overall design of the editor I must admit that I feel little upset too with the lack of review while Dennis was designing the whole scenes so we could come up with better ideas, and of course I'm completely open to change everything we need, don't get me wrong. The point is that he's still available for few more days so we could discuss this things with him instead of trying to solve just directly by overselves, so he could give us his point of view.

This discussions are very time consuming for the real benefits that we could get, so we could just try to move on with the features that we have hard time to agree and keep focusing on the bugs and stuff that we really need to have for MVP and after that we could keep collecting feedback from real users.

caseyyee:
Going to work on breaking this up into smaller pieces.   We can agree to disagree in smaller chunks.

PR #93, opened 2016-07-05
es6

ngokevin:
Do what all the cool kids are doing.
- ES6 class syntax to later support higher-order components.
- Use stateless functional components when state/lifecycle methods aren't needed.
- Use ES6 `export`

Tested:
- Creating entities
- Gizmos and helpers
- Modifying component props of all sorts
- Poking around the scene graph
- Removing entities
- Deleting components
- Adding components
- Reset components to default
- Assets modal
- Export HTML
- Enter / exit

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "devDependencies": {
     "babel-core": "^6.10.4",
     "babel-loader": "^6.2.4",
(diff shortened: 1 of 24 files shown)


fernandojsg:
Wow, quite huge and helpful amount of work here, thanks for doing it! 👍

PR #105, opened 2016-07-06
Remove ambiguous bullets

caseyyee:
image

Not sure how you all feel about this one.  I know the icons were contentious, but I certainly feel that the bullets don't really give any hints to what I am looking at.


fernandojsg:
We could just remove and in the future we'll add the arrows to collapse/expand children

PR #106, opened 2016-07-07
Fixes whitespace

caseyyee:
Also fixes some comments and removes commented/unused styles.


ngokevin:
@fernandojsg Gonna merge this through since we have multiple CSS patches in the works.

Issue #29, opened 2016-06-30
Number fields are confusing; use input[type=number]/input[type=range] instead of custom sliders


cvan:
![image](https://cloud.githubusercontent.com/assets/203725/16507979/1d7ae758-3ee3-11e6-8f98-c54c72317dce.png)

It took me a good minute or so to discover that by horizontally dragging in a numeric text field that the value changes. It's neat. But wasn't discoverable. Perhaps make that more obvious or require the field to be focussed first (that's what Dev Tools do, anyway). And possibly consider using the native browser ones: `input[type=number]`, `input[type=range]`.

cvan:
With `input[type=number]`:

![image](https://cloud.githubusercontent.com/assets/203725/16508026/88399ab2-3ee3-11e6-9128-ddae4afdea2e.png)

Which also gives you free up/down keyboard navigation - which is super powerful (and something familiar to those using similar tools).

PR #119, opened 2016-07-07
add visible borders to number/string inputs (related to issue #29)

cvan:
## before

before
## after

after

### DIFF
--- a/src/css/dark.css
+++ b/src/css/dark.css
@@ -27,25 +27,38 @@ input {
   color: #fff;
 }
 
+input[type="text"],
 input.string {
   outline: none;
 }
 
+input[type="number"],
 input.number {
-  color: #20b1fb !important;
-  font-size: 12px;
-  border: 0px;
-  padding: 2px;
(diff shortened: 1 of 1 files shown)


cvan:
this is just a tweak. to completely close #29, all numeric fields should be properly changed from `input.number` to `<input type="number">`/`<input type="range">`

fernandojsg:
Do you think we should center or align right the numbers with the new style?

caseyyee:
I think having a box around the numbers helps make it easy to see that it's a editable field.  Also helps with the readability of the numbers as well.   In this case, it's aligned-left:

image

(Sketch)

We can use underlines in place of boxes.  As long as we are consistent.

Also like that they have X, Y clearly labeled for the fields.

cvan:
👍 to Sketch's treatment

ngokevin:
https://github.com/aframevr/aframe-inspector/commit/ecb20458259544ac08925109d2ed557fb72581eb

PR #150, opened 2016-07-07
add minimum tolerance for number input dragging (fixes #42)

### LINKED ISSUE #42: Clicking on inputs can trigger accidentally rewrite values (due to horizontal stepping)
I just clicked on the field for the rotation X value, and it changed to `-0.02` because I must've dragged my mouse a bit on `mousedown`.

image

ngokevin:
Change at: https://github.com/aframevr/aframe-editor/compare/master...ngokevin:numtol?expand=1#diff-dc291811480fae652eda7cd59e52b139R47

### DIFF
--- a/src/components/widgets/NumberWidget.js
+++ b/src/components/widgets/NumberWidget.js
@@ -23,29 +23,34 @@ export default class NumberWidget extends React.Component {
 
   constructor (props) {
     super(props);
-    this.state = {value: this.props.value, displayValue: this.props.value.toFixed(this.props.precision)};
+    this.state = {
+      value: this.props.value,
+      displayValue: this.props.value.toFixed(this.props.precision)
+    };
   }
 
   componentDidMount () {
     this.distance = 0;
     this.onMouseDownValue = 0;
-    this.prevPointer = [ 0, 0 ];
+    this.prevPointer = [0, 0];
 
     this.setValue(this.props.value);
-
     this.onBlur();
   }
 
   onMouseMove = event => {
-    var currentValue = parseFloat(this.value);
-    var pointer = [ event.clientX, event.clientY ];
+    const currentValue = parseFloat(this.value);
+    const pointer = [event.clientX, event.clientY];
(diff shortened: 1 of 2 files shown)


src/components/widgets/NumberWidget.js
```diff
@@ -120,12 +130,12 @@ export default class NumberWidget extends React.Component {
   render () {
     return (
       <input ref='input' className='number' type='text'
-        value={this.state.displayValue}
-        onKeyDown={this.onKeyDown}
-        onChange={this.onChange}
-        onMouseDown={this.onMouseDown}
-        onFocus={this.onFocus}
-        onBlur={this.onBlur}
+             value={this.state.displayValue}
```
dmarcos: Is this indentation correct?

ngokevin: The default React linter doesn't have indent rules enabled. I've usually indented to line up the props. But by 2s is accepted as well.

dmarcos: What do you mean accepted as well? There should be just one way to indent. Just pick one.

ngokevin: i'll pick to line up visually with the first prop then

dmarcos: No. it's not a criteria you can consistently apply everywhere. Just use two spaces. You start to have context dependent indentation if you go that route. There's zero value on it.

dmarcos: It's @fernandojsg code base so I let him make the call

fernandojsg: I don't really have an strong opinion about the indent, but I would prefer to be consistent everywhere and not context dependent yep.
@ngokevin can we enable some indent rules in the linter even if by default they're disabled?

ngokevin: yeah

fernandojsg: Perfect, thanks!

PR #151, opened 2016-07-07
use number input for number widget (fixes #29)

### LINKED ISSUE #29: Number fields are confusing; use input[type=number]/input[type=range] instead of custom sliders

ngokevin:
- Change immediately `onChange` again.
- Enables arrow keys for number inputs.
- But don't show the arrows.

### DIFF
--- a/src/components/widgets/NumberWidget.js
+++ b/src/components/widgets/NumberWidget.js
@@ -106,7 +106,9 @@ export default class NumberWidget extends React.Component {
   }
 
   onChange = e => {
-    this.setState({value: e.target.value, displayValue: e.target.value});
+    const value = e.target.value;
+    this.setState({value: value, displayValue: value});
+    this.setValue(value);
   }
 
   onKeyDown = event => {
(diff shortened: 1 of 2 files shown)


fernandojsg:
I doesn't works on the input with decimals

src/components/widgets/NumberWidget.js
```diff
@@ -106,7 +106,9 @@ export default class NumberWidget extends React.Component {
   }
 
   onChange = e => {
-    this.setState({value: e.target.value, displayValue: e.target.value});
+    const value = e.target.value;
+    this.setState({value: value, displayValue: value});
+    this.setValue(value);
```
fernandojsg, line 111: If we set the value here it will set the format to `.toFixed` so it won't let you keep typing. If we just let the user write and it will just take the actual value onBlur (or onKeyPress enter) and just leave the onChange to change the display value.

ngokevin, line 111: OK. Looking to find a way to let it change `onChange` so you don't have to keep blurring to see the changes.

PR #173, opened 2016-07-08
Add transforms toolbar and make panel titles much smaller

fernandojsg:
Split fro https://github.com/aframevr/aframe-editor/pull/148

Added gizmos and support for local and global transforms.
screenshot 2016-07-08 20 09 30

Video of the gizmos working: https://www.dropbox.com/s/zck3dfwqyrunwto/transforms_gizmos.mov?dl=0

PS: The `transformControls` file doesn't need review directly as I didn't change anything there, it's taken  from the latest threejs version.

### DIFF
--- a/src/components/ToolBar.js
+++ b/src/components/ToolBar.js
@@ -1,0 +1,58 @@
+var React = require('react');
+var Events = require('../lib/Events.js');
+var classNames = require('classnames');
+
+var TransformButtons = [
+  {value: 'translate', icon: 'fa-arrows'},
+  {value: 'rotate', icon: 'fa-repeat'},
+  {value: 'scale', icon: 'fa-expand'}
+];
+
+export default class ToolBar extends React.Component {
+  constructor (props) {
+    super(props);
+    this.state = {selectedTransform: 'translate', localSpace: false};
+  }
+
+  changeTransformMode = mode => {
+    this.setState({selectedTransform: mode});
+    Events.emit('transformModeChanged', mode);
+  }
+
+  onLocalChange = e => {
+    var local = e.target.checked;
+    this.setState({localSpace: local});
+    Events.emit('spaceChanged', local ? 'local' : 'world');
+  }
+
+  renderTransformButtons = () => {
+    return TransformButtons.map(function (option) {
+      var selected = option.value === this.state.selectedTransform;
+      var classes = classNames({
+        button: true,
+        fa: true,
+        [option.icon]: true,
+        active: selected
+      });
+
+      return <a href='#' title={option.value}
+        onClick={this.changeTransformMode.bind(this, option.value)}
+        className={classes}></a>;
+    }.bind(this));
+  }
+
+  render () {
+    return (
+      <div className='toolbar'>
+        {this.renderTransformButtons()}
+        <span className='local-transform'>
+          <input id='local' type='checkbox'
+            checked={this.state.localSpace || this.state.selectedTransform === 'scale'}
+            disabled={this.state.selectedTransform === 'scale'}
+            onChange={this.onLocalChange}/>
+          <label htmlFor='local'>local</label>
+        </span>
+      </div>
+    );
+  }
+}
--- a/src/css/dark.css
+++ b/src/css/dark.css
@@ -235,29 +235,6 @@ input.number:focus {
     background-color: #111;
   }
 
-#toolbar {
-  position: absolute;
-  left: 0;
-  right: 330px;
-  bottom: 0;
-  height: 32px;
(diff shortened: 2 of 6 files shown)


caseyyee:
I don't think the panel titles are needed.  They just take up space.  It's pretty obvious what it is.

Transform toolbar looks good.  I think we can tweak the placement a bit, but this is a good start.

The last tool is scale right?

caseyyee:
This is super awesome btw.   Much needed.

src/css/dark.css
```diff
@@ -458,7 +435,7 @@ a.button {
 }
 
 a.button:hover {
-  color: #1faaf2;
+  color: #1faaf2 !important;
```
ngokevin: What are we trying to override?

src/components/ToolBar.js
```diff
@@ -0,0 +1,46 @@
...
+
+  onLocalChange = e => {
+    var local = e.target.checked;
+    this.setState({localSpace: local});
+    Events.emit('spaceChanged', local ? 'local' : 'world');
+  }
+
+  render () {
+    return <div className='toolbar'>
+      {
```
ngokevin: I usually try to keep code blocks out of the render to be more readable.

``` js
renderTransformButtons = () => {
  return TransformButtons.map  // ...
}

// ...

{this.renderTransformButtons()}
```

The fat arrow will also do the `this` binding for you.

src/components/ToolBar.js
```diff
@@ -0,0 +1,46 @@
...
+        TransformButtons.map(function (option) {
+          var selected = option.value === this.state.selectedTransform;
+          return <a href='#' title={option.value}
+            onClick={this.changeTransformMode.bind(this, option.value)}
+            className={'button fa ' + option.icon + (selected ? ' active' : '')}></a>;
```
ngokevin: We could use https://www.npmjs.com/package/classnames

``` js
var classes = {
  button: true,
  fa: true,
  [option.icon]: true,
  active: selected
};
//...
className={...classes}
```

src/css/dark.css
ngokevin: Try to avoid `!important` as it tends to spread, see if we can just add more specificity. (you can repeat class name to add more if you need `a.button.button:hover`

src/components/ToolBar.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  onLocalChange = e => {
+    var local = e.target.checked;
+    this.setState({localSpace: local});
+    Events.emit('spaceChanged', local ? 'local' : 'world');
+  }
+
+  render () {
+    return <div className='toolbar'>
```
ngokevin: I like wrapping to separate JSX:

``` js
return (
  <div>
  // ...
  </div>
);
```

ngokevin:
Code looks good. Will let Casey handle the UX bits.

src/css/dark.css
fernandojsg: I don't really know why I added this :D I'm just removing it in the new commit.

ngokevin: Check the other `!important` in this file too

caseyyee:
If this includes updates to the toolbar aligned left of attributes panel (as discussed on slack), we're good :)

src/components/ToolBar.js
fernandojsg: I've added that, but do you mean this way right?

``` javascript
      var classes = classNames({
        button: true,
        fa: true,
        [option.icon]: true,
        active: selected
      });

      return <a href='#' title={option.value}
        onClick={this.changeTransformMode.bind(this, option.value)}
        className={classes}></a>;
```

src/components/ToolBar.js
fernandojsg: Yep

src/components/ToolBar.js
ngokevin: Yeah :)

fernandojsg:
@caseyyee yep, I've added the changes we discussed on slack. 
screenshot 2016-07-08 20 56 24

Regarding the icon the last one is the scale, until we get a better icons maybe we could use this one -> http://fontawesome.io/icon/arrows-alt/ ?

src/components/ToolBar.js
fernandojsg: Thanks! I've done it too

caseyyee:
That is the "expand" icon.   I think the one you have up there is better.   I couldn't find anything on FA that had something similar to: 

image

what you have for now will do.

fernandojsg:
Ok! So I'm merging it 
thank you!

PR #177, opened 2016-07-08
Fixed #110 add debug=true to a-scene and call flushToDOM

### LINKED ISSUE #110: Set <a-scene> debug to true and call flushToDOM
so component changes are reflected on the DOM

fernandojsg:
Set the `debug=true` once the scene is loaded and call `flushToDOM()` to refresh the DOM.

### DIFF
--- a/src/lib/editor.js
+++ b/src/lib/editor.js
@@ -30,6 +30,10 @@ Editor.prototype = {
   },
 
   onSceneLoaded: function () {
+    this.previousDebug = this.sceneEl.hasAttribute('debug') && this.sceneEl.hasAttribute('debug') !== 'false';
+    this.sceneEl.setAttribute('debug', 'true');
+    this.sceneEl.flushToDOM();
+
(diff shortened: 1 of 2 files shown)


src/lib/editor.js
```diff
@@ -30,6 +30,9 @@ Editor.prototype = {
   },
 
   onSceneLoaded: function () {
+    this.sceneEl.setAttribute('debug', 'true');
```
ngokevin, line 34: What about removing it once you exit the Editor if it wasn't already set?

fernandojsg, line 34: Yep, good point. I did this, and it's working but it doesn't reflect the changes on the DOM (It keeps showing `debug=true` but acting as `false`. It's like the `flushToDOM` doesn't do anything

``` javascript
    this.sceneEl.setAttribute('debug', this.previousDebug ? 'true' : 'false');
    this.sceneEl.flushToDOM();
```

PR #179, opened 2016-07-08
fix vec3 widget (fixes #176)

### LINKED ISSUE #176: Vec3 doesn't reflect changes when componentChanged
When you modify the `position`, `rotation` or `scale` of an object using the transforms gizmos or directly from the DOM inspector, they don't change the value of their `vec3` widgets.
This is something we just missed on the latest changes of the `NumericWidget` as it was working properly in the early versions. /cc @ngokevin

ngokevin:
Need to account for updating props and combining into state.


fernandojsg:
superb! thanks

PR #181, opened 2016-07-09
Fixes export and copy to clipbard scene issues

fernandojsg:
Delete everything injected by `aframe` or `aframe-editor` when exporting to HTML or copying to clipboard the entire scene.

### DIFF
--- a/src/lib/exporter.js
+++ b/src/lib/exporter.js
@@ -16,9 +16,25 @@ module.exports = {
 
     // Remove all the components that are being injected by aframe-editor or aframe
     // @todo Use custom class to prevent this hack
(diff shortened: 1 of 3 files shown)


src/lib/exporter.js
```diff
@@ -16,7 +16,21 @@ module.exports = {
 
     // Remove all the components that are being injected by aframe-editor or aframe
     // @todo Use custom class to prevent this hack
-    Array.prototype.forEach.call(xmlDoc.querySelectorAll('.a-enter-vr,.a-orientation-modal,.Panel,.aframe-editor,.rs-base,.a-canvas,.a-enter-vr-button,style[data-href="style/rStats.css"],style[data-href^="src/panels"],style[data-href="style/aframe-core.css"],link[href^="https://maxcdn.bootstrapcdn.com"]'), function (el) {
+    Array.prototype.forEach.call(xmlDoc.querySelectorAll([
```
ngokevin: i guess this way is faster than a `for loop`?

also does the `Array.prototype.forEach` on the `NodeList` work okay on firefox?

ngokevin:
r+wc

src/lib/exporter.js
fernandojsg: Actually I believe `forEach` should be slower than `for loop` but as It should be a very big amount of data probably It won't make a difference. I tried it too on firefox and seems to works.
In any case, I've changed it to use `for loop` just to keep it simple.

PR #188, opened 2016-07-10
Fixes #56 Copy entity HTML to clipboard

### LINKED ISSUE #56: Allow copy markup of a single entity
Imagine I'm inspecting a scene to learn how it works. I see an entity with a combination of components that does what I need. It would be nice to be able to copy and paste it in my editor. We could add a button on a contextual menu of the entity in the 3D view or/and a button on the entity inspector on the right.

fernandojsg:
Added copy to clipboard icon on the entity panel. It will copy the HTML for the current component and its children.
screenshot 2016-07-10 15 05 41

Icons and style _to-be-defined_ to match the same style for the component's icons (https://github.com/aframevr/aframe-editor/issues/10)


dmarcos:
If the copy icon is on common it seems that you're going to only copy the common section

dmarcos:
The icon / label should be better placed on the entity on the scenegraph

ngokevin:
Advantage on the component panel is to keep the entity copy action in the same place as component copy action. As well as separating management of scene structure (scene graph) from scene data (components panel).

But I agree the current placement makes it seem like you can only copy the common components.

src/components/attributes/CommonComponents.js
```diff
@@ -7,6 +7,8 @@ import Collapsible from '../Collapsible';
 import MixinsComponent from './MixinsComponent';
 import {updateEntity} from '../../actions/entity';
 
+var Clipboard = require('clipboard');
```
ngokevin: try `import Clipboard from 'clipboard';`

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
-  return (
-    <Collapsible>
-      <div className='collapsible-header'>
-        <span>Common</span>
-      </div>
-      <div className='collapsible-content'>
-        <div className='row'>
-          <span className='value tagName'><code>&lt;{entity.tagName.toLowerCase()}&gt;</code></span>
+class CommonComponents extends React.Component {
+
```
ngokevin: You can move the propTypes up here now that it's not stateless

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
+  componentDidMount () {
+    var self = this;
+    var clipboard = new Clipboard('[data-action="copy-entity-to-clipboard"]', {
+      text: function (trigger) {
```
ngokevin: if you use es6 arrow, you get automatic binding

``` js
text: trigger => {
  return this.props...
}
```

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
+        return self.props.entity.outerHTML;
+      }
+    });
+    clipboard.on('error', function (e) {
```
ngokevin: here as well, then you don't need `self`

```
e => {

}
```

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
+      // @todo Show the error on the UI
+    });
+  }
+
+  render () {
+    const entity = this.props.entity;
+    const components = entity ? this.props.entity.components : {};
```
ngokevin: you can use `entity` instead of `this.props.entity`

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
+          <div className='row'>
+            <span className='value tagName'><code>&lt;{entity.tagName.toLowerCase()}&gt;</code></span>
+          </div>
+          <div className='row'>
+            <span className='text'>ID</span>
+            <InputWidget onChange={changeId} entity={entity} name='id' value={entity.id}/>
+          </div>
+          {
+            Object.keys(components).filter(function (key) {
+              return ['visible', 'position', 'scale', 'rotation'].indexOf(key) !== -1;
```
ngokevin: These components are also defined in `Attributes.js`, perhaps move them to a constants or share them.

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
+            }).map((key) => {
+              const componentData = components[key];
+              const schema = AFRAME.components[key].schema;
+              return (
+                <AttributeRow onChange={updateEntity} key={key} name={key}
+                  schema={schema} data={componentData.data} componentname={key}
+                  entity={this.props.entity}/>
```
ngokevin: `entity` works

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
-          <InputWidget onChange={changeId} entity={entity} name='id' value={entity.id}/>
+        <div className='collapsible-content'>
+          <div className='row'>
+            <span className='value tagName'><code>&lt;{entity.tagName.toLowerCase()}&gt;</code></span>
+          </div>
+          <div className='row'>
+            <span className='text'>ID</span>
+            <InputWidget onChange={changeId} entity={entity} name='id' value={entity.id}/>
+          </div>
+          {
```
ngokevin: could abstract this codeblock out to a method

src/css/dark.css
```diff
@@ -320,6 +320,10 @@ div.vec3 {
   width: 50px;
 }
 
+.collapsible-header {
+  display: flex;
```
ngokevin: nice

src/components/attributes/CommonComponents.js
```diff
@@ -15,41 +17,61 @@ function changeId (entity, componentName, propertyName, value) {
...
+            Object.keys(components).filter(function (key) {
+              return ['visible', 'position', 'scale', 'rotation'].indexOf(key) !== -1;
+            }).map((key) => {
```
ngokevin: `componentName` would be more descriptive

you can also do `key =>` without the parens if it's just one arg

src/components/attributes/CommonComponents.js
fernandojsg: done

src/components/attributes/CommonComponents.js
fernandojsg: done

src/components/attributes/CommonComponents.js
fernandojsg: done

src/components/attributes/CommonComponents.js
fernandojsg: done

src/components/attributes/CommonComponents.js
fernandojsg: done

src/components/attributes/CommonComponents.js
fernandojsg: What do you propose? maybe having a constant defined on CommonAttributes and export it from there? Or just a different file with just the constants

src/components/attributes/CommonComponents.js
fernandojsg: yep thank u

src/components/attributes/CommonComponents.js
fernandojsg: done => `renderCommonAttributes`

src/components/attributes/CommonComponents.js
fernandojsg: done

src/components/attributes/CommonComponents.js
ngokevin: Maybe just in Common...since it's only one constant for now.

fernandojsg: Ok, done

ngokevin:
Code looks good. Just need to address the UX.
- Current icon placement looks like it applies for common components.

Possible solutions:
- Add text label to the icon. Possibly renaming the Common panel to the Entity panel might help as well.
- Move function to scene graph - I don't quite like this as it mixes scene structure with entity data.

dmarcos:
Renaming `common` to `entity` is misleading. All the components below belong to the entity as well. A third solution:

We could make the `attributes` panel title (and maybe rename to `entity`) larger to accommodate the copy button. That would break the consistency though with the scenegraph title.

dmarcos:
As a side note. I think we still need to visually differentiate the commons section from the rest of the components. Commons looks like any other component

dmarcos:
A fourth solution: We can put the copy button next to the `<a-entity>` label

ngokevin:
Third solution addendum, we can make both headers bigger. They do look sort of cramped anyways.

fernandojsg:
I like the third solution too, I'll make a mockup and we could discuss about it.

fernandojsg:
What do you think about something like this (Adding the copy to clipboard icon as the third proposed solution by @dmarcos)?
I was thinking that probably we could remove the `<a-entity>` from common and put it in the panel title, so instead of having, like in the screenshot, `entity`we could have `<a-entity>`, `<a-box>`, `<a-entity #box>`...
screenshot 2016-07-15 20 26 45

dmarcos:
The top padding / margin of the titles looks too dark in comparison with the lines that separate the component titles. They should be consistent. component title

fernandojsg:
@dmarcos I just fixed that, although the menu bar will be remove in the next PR :D

fernandojsg:
Updated with the last changes regarding the top menu. I've moved the `<a-entity>` attribute from `Common` to the title of the sidebar.
screenshot 2016-08-01 18 42 00

fernandojsg:
@dmarcos @ngokevin what do you think about the new style? ^

dmarcos:
Maybe the question mark for the docs is too big. We sould make it match the line height of the component title

dmarcos:
Should we make `<a-entity>` title larger and increase the contrast? We probably need to emphasize that it's the title for the whole components panel

fernandojsg:
@dmarcos yep, I'll decrease the size of the icon for help.
Re the title, I've made it white and a little bigger, something like this you mean?
screenshot 2016-08-03 01 47 09

fernandojsg:
Maybe a different color of the title background to make a bigger contrast?

fernandojsg:
@dmarcos @ngokevin what do you think about adding this functionality as is and then we could create another PR just for style the icons and sidebar title style?

src/components/components/ComponentsContainer.js
```diff
@@ -11,34 +10,21 @@ export default class ComponentsContainer extends React.Component {
...
   };
 
-  constructor (props) {
-    super(props);
-    this.state = {
-      entity: props.entity
-    };
-  }
-
+/*
```
ngokevin: could remove

src/css/main.css
```diff
@@ -878,11 +878,21 @@ code, pre {
 }
 
 .sidebar-title {
+/*
```
ngokevin: could remove

ngokevin:
I get: `<a-entity id="greenBox" geometry="primitive: box" material="color: #0f0" position="4 2 4" rotation="" scale="" visible=""></a-entity>`

Should we trim the null values for default components?

ngokevin:
r+wc adjusting style in another PR sounds good.

fernandojsg:
@ngokevin do you mean converting this 

```
<a-entity id="greenBox" geometry="primitive: box" material="color: #0f0" position="4 2 4" rotation="" scale="" visible=""></a-entity>
```

into

```
<a-entity id="greenBox" geometry="primitive: box" material="color: #0f0" position="4 2 4" rotation scale visible></a-entity>
```

or directly removing them as they're already default components included in any entity?

```
<a-entity id="greenBox" geometry="primitive: box" material="color: #0f0" position="4 2 4"></a-entity>
```

ngokevin:
I would polish it so it is ready to paste into people's code editors, meaning to remove superfluous stuff that don't matter like `rotation=""`. So yeah, I'd say remove them.

fernandojsg:
@ngokevin I've implemented removing the default attributes if they've default value, thanks for the idea. the result copy to the clipboard it's more clear now.

PR #189, opened 2016-07-10
Added icons and remove context menu on the components bar. Fixes #10

### LINKED ISSUE #10: The command menu of the component should appear below the three dots
I clicked on the three dots icon and remove the component by mistake. The non reversible operations should be harder to trigger to avoid unintented results.

component menu

fernandojsg:
- Removed the three dots and the contextmenu on the component's name and replace them by icons (Style and image _to-be-defined_)
- Added confirmation while trying to remove or reset a component.

screenshot 2016-07-10 15 18 50

### DIFF
--- a/src/components/components/Component.js
+++ b/src/components/components/Component.js
@@ -20,14 +20,28 @@ export default class Component extends React.Component {
     name: React.PropTypes.string
   };
 
-  deleteComponent = event => {
-    event.stopPropagation();
-    this.props.entity.removeAttribute(this.props.name);
+  constructor (props) {
+    super(props);
+    this.state = {
+      entity: this.props.entity,
+      name: this.props.name
+    };
   }
 
-  resetComponent = event => {
+  componentWillReceiveProps (newProps) {
+    if (this.state.entity !== newProps.entity) {
+      this.setState({entity: newProps.entity});
+    }
+    if (this.state.name !== newProps.name) {
+      this.setState({name: newProps.name});
+    }
+  }
+
+  removeComponent = event => {
     event.stopPropagation();
-    this.props.entity.setAttribute(this.props.name, {});
+    if (confirm('Do you really want to remove component `' + this.props.name + '`?')) {
+      this.props.entity.removeAttribute(this.props.name);
+    }
   }
 
   render () {
@@ -63,12 +77,9 @@ export default class Component extends React.Component {
       <Collapsible>
         <div className='collapsible-header'>
           <span title={subComponentName || componentName}>{subComponentName || componentName}</span>
-          <div className='dropdown menu'>
-            <div className='dropdown-content'>
-              <a href='#' onClick={this.deleteComponent}>Delete</a>
-              <a href='#' onClick={this.resetComponent}>Reset to default</a>
-              <a href='#' className='disabled'>Copy to clipboard</a>
-            </div>
+          <div>
+            <a href='#' title='Remove component' className='flat-button'
+              onClick={this.removeComponent}>remove</a>
           </div>
         </div>
         <div className='collapsible-content'>{propertyRows}</div>
--- a/src/components/modals/Modal.js
+++ b/src/components/modals/Modal.js
@@ -10,7 +10,7 @@ export default class Modal extends React.Component {
     document.addEventListener('keyup', this.handleGlobalKeydown);
     document.addEventListener('mousedown', this.handleGlobalMousedown);
   }
-
+change
   handleGlobalKeydown = event => {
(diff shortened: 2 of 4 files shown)


dmarcos:
Where is the icon to copy the component to clipboard going to be?

fernandojsg:
@dmarcos next to the current icons, in the same row

dmarcos:
Are we adding the icon on a different PR or should be part of this one?

fernandojsg:
Not yet, the icons will be something like the one used on the Copy to clipboard for entity (https://github.com/aframevr/aframe-editor/pull/188) and it will be added in another PR, not to mix different issues in the same PR (delete dots -> icons, and copy to clipboard)

dmarcos:
Can we try labels instead of icons as @cvan suggested? It might not be too bad.

fernandojsg:
My idea is, as they all looks the same style, review the functionality and the placement and if we're good to go merge them and create another issue with "Style the component's icons"

fernandojsg:
do you mean label when mouseOver or label all the time?

dmarcos:
Label all the time: `copy`, `reset`, `remove` instead of the icon.

dmarcos:
We can add a `subtle background` to indicate it's a button

fernandojsg:
You mean something like this right?


(From https://github.com/aframevr/aframe-editor/issues/10)

fernandojsg:
I'm afraid that showing the labels all the time won't fit in some situations, like when we've multiple components:
screenshot 2016-06-29 01 57 11

dmarcos:
When having multiple components we can show the id in the panel like the html element (`<a-entity>`) in the common section

fernandojsg:
But could be nice to see them when everything is collapsed

dmarcos:
What about always showing the attribute name as it is. Only `SOUND__2` instead of `SOUND (SOUND__2)` ?

fernandojsg:
We had already an issue on the previous repo but I deleted when I changed the repos, I've created a new one to discuss it again -> https://github.com/aframevr/aframe-editor/issues/190

ngokevin:
Use `text-overflow: ellipsis` + `max-width` up until the start of the icons so they don't clash. Then use `title` so they can see the full name on hover.

`SOMESUPERLONGCOMPONENT__TRUNC...    icon1 icon2`

fernandojsg:
@ngokevin ok thank you! And what do you think about the icons + tooltip/label? https://github.com/aframevr/aframe-editor/issues/10

ngokevin:
The current state is great. Only problems:
- Tooltip takes too long to display on Chrome after hover. We can't really change that except by using a custom solution.
- Eraser icon might not be immediately clear what it does. The alert modal does help. Personally, I don't think we need Reset to Default and would remove it (Trashing + Re-add would do it too). But if we keep it, we'll have telemetry to see how often it is used.

Addressing those two might not be in context of this PR.

dmarcos:
What about using labels instead of icons? text labels with a `background` + underline to convey clickability ( [`remove`](), [`delete`]() )

ngokevin:
That sounds good, it would solve both issues.

fernandojsg:
Something like this?
screenshot 2016-07-12 00 30 49
I also start to believe that the `clear` button probably won't be used so much, so we could remove it and just add it in the future if someone asks for it. Or just leave it on the `edit` menu, just in case :)

ngokevin:
That looks good. Maybe lowercase `copy html` too.

fernandojsg:
I've removed the `reset to default` button and the `copy html` because it was just to show the mockup and I'll open another PR with that functionality.
screenshot 2016-07-12 00 57 28

src/components/attributes/Component.js
```diff
@@ -60,12 +74,9 @@ export default class Component extends React.Component {
...
         <div className='collapsible-header'>
           <span title={subComponentName || componentName}>{subComponentName || componentName}</span>
-          <div className='dropdown menu'>
-            <div className='dropdown-content'>
-              <a href='#' onClick={this.deleteComponent}>Delete</a>
-              <a href='#' onClick={this.resetComponent}>Reset to default</a>
-              <a href='#' className='disabled'>Copy to clipboard</a>
-            </div>
+          <div>
+            <a href="#" title="Delete component" className="flat-button"
```
ngokevin: single quotes

ngokevin:
`remove` component is more consistent to the API than `delete`. Former means to detach from the entity, latter almost sounds like to permanently delete.

r+wc love it

fernandojsg:
@ngokevin both changes done! thank you

PR #193, opened 2016-07-11
	small refactor to component property rendering

ngokevin:
<s>Ad-hoc solution for addressing a common complaint. These few properties won't do anything visual from the Editor. So remove them to clean up the noise on one of the more important components.</s>

Decision changed. Only left the refactor.

### DIFF
--- a/src/components/components/Component.js
+++ b/src/components/components/Component.js
@@ -1,14 +1,9 @@
+/* global AFRAME */
 import React from 'react';
-
 import PropertyRow from './PropertyRow';
 import Collapsible from '../Collapsible';
 
-function isSingleProperty (schema) {
-  if ('type' in schema) {
-    return typeof schema.type === 'string';
-  }
-  return 'default' in schema;
-}
+const isSingleProperty = AFRAME.schema.isSingleProperty;
 
 /**
  * Single component.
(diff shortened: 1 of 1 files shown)


dmarcos:
I'm not sure if it's good going down that path. What's the criteria to hide properties? There are just a few properties that are noisy and even in that case the fact that show up in the editor makes them discoverable. What about this? Let's make an `editable` boolean in the property schema. If editable is false the editor still shows the property but grayed out. It shows the value but it's not modifiable.

ngokevin:
Yeah, it's very ad-hoc. These are the only three offending properties at the moment. I think this is okay to polish the initial release and then find a scalable solution later (i.e,., allowing components to configure editor appearance).

dmarcos:
Can we make them non editable instead of hiding them? At least you can see the property and its current value.

ngokevin:
The issue was more that they are visible in the first place.

So it's usability (modifying relevant properties that have any actual effect in the Editor) vs. property discoverability (seeing all the properties). I think disabling them would help, so people don't try to modify them and see nothing happen, though it'd be slightly weird to expose non-editable properties in the Editor.

What about https://github.com/aframevr/aframe-editor/issues/31 for property discoverability? Since at least component discoverability is not affected by this.

dmarcos:
It deviates the editor from its purpose of teaching you about components. I don't think that seeing those properties is that big of an issue. They're going to be a tiny fraction of them.

dmarcos:
Even if those properties are not editable at runtime it can be useful to see their value to diagnose problems. Something could be affecting the appearance of the entities and one of those hidden properties could be the reason. I believe in the hide nothing approach.

fernandojsg:
I also like the approach of disable them but put everything visible.

ngokevin:
Fair enough. I have repurposed the PR to just the refactor changes.

ngokevin:
Rebased.

PR #196, opened 2016-07-12
Copy component to clipboard

fernandojsg:
Added copy component data to clipboard:
screenshot 2016-07-12 02 39 12

It just copy the values modified from the default schema. For the previous image the string copied to the clipboard is:
`color: #fafafa; metalness: 0.2; repeat: 50 20; roughness: 0.1; src: #floorImg`

### DIFF
--- a/src/actions/component.js
+++ b/src/actions/component.js
@@ -1,0 +1,37 @@
+/**
+ * Get the list of modified properties
+ * @param  {Element} entity        Entity where the component belongs
+ * @param  {string} componentName Component name
+ * @return {object}               List of modified properties with their value
+ */
+function getModifiedProperties (entity, componentName) {
+  var data = entity.components[componentName].data;
+  var defaultData = entity.components[componentName].schema;
+  var diff = {};
+  for (var key in data) {
+    var defaultValue = defaultData[key].default;
+    var currentValue = data[key];
+
+    // Some parameters could be null and '' like mergeTo
+    if (!(isEmpty(currentValue) && isEmpty(defaultValue)) &&
+      currentValue !== defaultValue) {
+      diff[key] = data[key];
+    }
+  }
+  return diff;
+}
+
+/**
+ * [getClipboardRepresentation description]
+ * @param  {[type]} entity        [description]
+ * @param  {[type]} componentName [description]
+ * @return {[type]}               [description]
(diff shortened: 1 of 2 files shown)


fernandojsg:
Now I'm thinking about the name... `copy html` could be understandable in the `entity` level as you're actually copying the whole html element, but here that we're just copying the attributes, do you think it's still a good name?

ngokevin:
i think 'copy html' is okay given not many other alternatives and to match the action of the entity copy

fernandojsg:
@ngokevin ok!, and what about the code? it's good to go?

src/actions/component.js
```diff
@@ -0,0 +1,48 @@
...
+  return diff;
+}
+
+/**
+ * Returns an string representation of an attributes list
+ * @param  {object} attributes List of attributes
+ * @return {string}            Text representation: 'at1: value; at2: value;'
+ */
+function attributesToText(attributes) {
+  return Object.keys(attributes).map(function(k) {
```
ngokevin: You can use `AFRAME.utils.styleParser.stringify`

src/actions/component.js
```diff
@@ -0,0 +1,48 @@
+/**
```
ngokevin, line 1: You can `npm run lintfile src/actions/component.js`

src/actions/component.js
```diff
@@ -0,0 +1,48 @@
...
+ * Get the list of modified attributes
+ * @param  {Element} entity        Entity where the component belongs
+ * @param  {string} componentName Component name
+ * @return {object}               List of modified attributes with their value
+ */
+function getModifiedAttributes(entity, componentName)  {
```
ngokevin: `s/attributes/properties`

ngokevin:
r+wc

src/actions/component.js
fernandojsg: Cool, I didn't know that :)

src/actions/component.js
fernandojsg, line 1: Yep, done! sorry for that

src/actions/component.js
fernandojsg: done! :)

fernandojsg:
thank you!

PR #197, opened 2016-07-12
Added clone and remove entity to the Edit menu

### LINKED ISSUE #36: Clone entity
It would be super useful to be able to clone a entity.

fernandojsg:
- Added `clone` and `remove` entity to the `Edit` menu
- ID Collision detection that will add an incremental suffix 
  screenshot 2016-07-12 17 41 00
- Remove entity ask for confirmation
- `Menu` => `ExportMenu`, `CreateMenu` and `EditMenu`
- Fixes https://github.com/aframevr/aframe-editor/issues/36 and https://github.com/aframevr/aframe-editor/issues/13

### DIFF
--- a/src/actions/entity.js
+++ b/src/actions/entity.js
@@ -47,11 +47,36 @@ export function updateEntity (entity, componentName, propertyName, value) {
 export function removeEntity (entity, force) {
   if (entity) {
     if (force === true || confirm('Do you really want to remove the entity: `' + (entity.id || entity.tagName) + '`')) {
+      var closest = findClosestEntity(entity);
       entity.parentNode.removeChild(entity);
-      // @todo Select the next entity in the scenegraph
-      editor.selectEntity(null);
+      editor.selectEntity(closest);
     }
   }
+}
+
+function findClosestEntity (entity) {
+  // First we try to find the after the entity
+  var nextEntity = entity.nextElementSibling;
+  while (nextEntity && (!nextEntity.isEntity || nextEntity.isEditor)) {
+    nextEntity = nextEntity.nextElementSibling;
+  }
+
+  // Return if we found it
+  if (nextEntity && nextEntity.isEntity && !nextEntity.isEditor) {
+    return nextEntity;
+  }
+  // Otherwise try to find before the entity
+  var prevEntity = entity.previousElementSibling;
+  while (prevEntity && (!prevEntity.isEntity || prevEntity.isEditor)) {
+    prevEntity = prevEntity.previousElementSibling;
+  }
+
+  // Return if we found it
+  if (prevEntity && prevEntity.isEntity && !prevEntity.isEditor) {
+    return prevEntity;
+  }
+
+  return null;
 }
 
 /**
@@ -60,6 +85,15 @@ export function removeEntity (entity, force) {
  */
 export function removeSelectedEntity (force) {
   removeEntity(editor.selectedEntity);
+}
+
+/**
+ * Insert an node after a referenced node.
+ * @param  {Element} newNode       Node to insert.
+ * @param  {Element} referenceNode Node used as reference to insert after it.
+*/
+function insertAfter (newNode, referenceNode) {
+  referenceNode.parentNode.insertBefore(newNode, referenceNode.nextSibling);
 }
 
 /**
@@ -73,9 +107,14 @@ export function cloneEntity (entity) {
   });
 
   // Get a valid unique ID for the entity
-  copy.id = getUniqueId(entity.id);
-  //insertAfter(copy, entity);
-  entity.insertAdjacentHTML('afterend', copy.outerHTML);
+  if (entity.id) {
+    copy.id = getUniqueId(entity.id);
+  }
+  copy.addEventListener('loaded', function () {
+    Events.emit('sceneModified');
+    editor.selectEntity(copy);
+  });
+  insertAfter(copy, entity);
 }
 
 /**
@@ -83,15 +122,6 @@ export function cloneEntity (entity) {
  */
 export function cloneSelectedEntity () {
   cloneEntity(editor.selectedEntity);
-}
-
-/**
- * Insert an node after a referenced node.
- * @param  {Element} newNode       Node to insert.
- * @param  {Element} referenceNode Node used as reference to insert after it.
- */
-function insertAfter (newNode, referenceNode) {
-  referenceNode.parentNode.insertBefore(newNode, referenceNode.nextSibling);
 }
 
 /**
--- a/src/actions/entity.js
+++ b/src/actions/entity.js
@@ -1,4 +1,4 @@
-/* global AFRAME */
+/* global AFRAME editor */
 var Events = require('../lib/Events.js');
 
 var components = AFRAME.components;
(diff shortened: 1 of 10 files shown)


src/lib/editor.js
```diff
@@ -83,6 +83,10 @@ Editor.prototype = {
       this.addObject(event.target.object3D);
     }.bind(this));
 
+    document.addEventListener('componentchanged', function(event) {
```
ngokevin, line 86: `event => { ... }` will give you the binding

src/components/menu/Menu.js
```diff
@@ -1,140 +1,20 @@
...
-
-    saveString(Exporter.generateHtml(), 'ascene.html');
-  }
-
   render() {
     return (
-      <div className="Panel" id="menubar">
-        <div className="menu aframe-logo">
-          <div className="title">
+      <div className='Panel' id='menubar'>
```
ngokevin: can we do lower case class names

src/components/menu/EditMenu.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const EditMenu = () => {
+  var prevGroup = null;
+  return (
+    <div className="menu">
+      <div className="title">Edit</div>
+      <div className="options">
+      {
+        Object.keys(menuOptions).map(function(key) {
```
ngokevin: can do `key =>` for binding

lint should raise a couple things for this file too

src/actions/entity.js
```diff
@@ -38,3 +38,79 @@ export function updateEntity (entity, componentName, propertyName, value) {
   }
   Events.emit('objectChanged', entity);
 }
+
```
ngokevin, line 41: can run linter

src/actions/entity.js
```diff
@@ -38,3 +38,79 @@ export function updateEntity (entity, componentName, propertyName, value) {
...
+    }
+  }
+}
+
+/**
+ * Remove the selected entity
+ * @param  {boolean} force (Optional) If true it won't ask for confirmation
+ */
+export function removeSelectedEntity (force) {
+  removeEntity(editor.selectedEntity);
```
ngokevin, line 87: can we pass in the entity rather than rely on global

src/actions/entity.js
```diff
@@ -38,3 +38,79 @@ export function updateEntity (entity, componentName, propertyName, value) {
...
+  cloneEntity(editor.selectedEntity);
+}
+
+/**
+ * Insert an node after a referenced node.
+ * @param  {Element} newNode       Node to insert.
+ * @param  {Element} referenceNode Node used as reference to insert after it.
+ */
+function insertAfter(newNode, referenceNode) {
+  referenceNode.parentNode.insertBefore(newNode, referenceNode.nextSibling);
```
ngokevin: https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML

src/actions/entity.js
```diff
@@ -38,3 +38,79 @@ export function updateEntity (entity, componentName, propertyName, value) {
...
+
+  var i = 2;
+  // If the baseId ends with _#, it extracts the baseId removing the suffix
+  var groups = baseId.match(/(\w+)_(\d+)/);
+  if (groups) {
+    baseId = groups[1];
+    i = groups[2];
+  }
+
+  while (document.getElementById(baseId + '_' + i)) { i++; }
```
ngokevin: how about hyphen instead of underscore

fernandojsg: yep, done!

src/components/menu/Menu.js
fernandojsg: done

src/components/menu/EditMenu.js
fernandojsg: done!

src/actions/entity.js
fernandojsg, line 41: :angel: sorry for that :) done

ngokevin:
Is the Edit Menu the best place? There's a lot of distance between the Entity in the SceneGraph to the Edit Menu, so the mapping is not immediate. Can we move these actions to the SceneGraph (right-click contextual menu or action icons)

src/actions/entity.js
fernandojsg: So should I do something like: 

``` javascript
entity.insertAdjacentHTML('afterend', copy.outerHTML);
```

Right?

fernandojsg:
@ngokevin regardless the position on the UI I think it's good idea to keep all the options in the menu too, so people could access them even if some of the panels are collapsed for example.
In any case I also think the best place to have the icons is also in the scenegraph, maybe you could click an entity and the delete button will be aligned to the right in the same row.
I also like the contentmenu option, probably even on the viewport too, so you could click on the entities and see some options related to them.
This PR is more about adding the functionality, and then I could add another two one for context menu and another for icons in scenegraph.

src/actions/entity.js
fernandojsg, line 87: Yep, in fact we have exposed both functions, but on the `EditMenu.js` I used the following code as the menu doesn't have a reference to the current selected entity:

``` javascript
var menuOptions = {
  'Clone': {group: 'base', callback: cloneSelectedEntity, needsEntity: true},
  'Remove': {group: 'base', callback: removeSelectedEntity, needsEntity: true}
};
```

Maybe `callback` could be the one calling `removeEntity(editor..')` and delete the `*selected` on `entity.js`

ngokevin:
We don't have collapsible panels at the moment, so right now it's not ideal to address that case. I think even then, we'd just prescribe to open up the Scene Graph to do entity-level management.

dmarcos:
What about a contextual menu (with clone, delete) as well on the viewport when right-click on an entity?

dmarcos:
If we add the delete, clone on both the viewport and the scene graph we might be able to get rid of the edit menu. Edit menu feels indirect while a contextual menu / icon feels more like direct manipulation which reinforces the mapping between action and target of that action.

fernandojsg:
Yep, I was more on the way to keep all the actions on the menus as most of the apps does, like firefox you've context menu to copy, paste, whatever but you have also shortcuts and at the end you can always go to the menu.
I believe for me the most confortable will be probably contextmenu on viewport and shortcuts but this is more for "power" users.
About adding icons to the scenegraph, probably delete it's the most common that you can have there, and maybe clone too?
What I mean is that there's no problem in duplicate places where you can do the same action so  isn't? so we could have them all in the menu

fernandojsg:
Btw from the discussion in https://github.com/aframevr/aframe-editor/pull/203
Do you think we should include the clone icon too in next to the delete icon on the selected entity?
screenshot 2016-07-14 00 28 16

dmarcos:
How do you create a child of a given entity? If you select one and click on the `+` icon?

dmarcos:
What about copy and paste for entities instead of clone? You copy an entity and can click on another one to paste it as a child. That would allow for easy creation of hierarchies.

fernandojsg:
Copy/paste should be on the contextmenu instead of icons right?
I'm not sure about to create a child of the selected entity with just pressing `+`. 
Because typically you'll have always one entity selected (instead of `a-scene`) Unity/3DS and so on typically works that way, you just add an entity at the end of the scenegraph and then you decide where to locate it.
One option could be having two `+` buttons one to add an entity at the end of the scenegraph and another one (just enabled if one entity is selected) to add an entity inside the current selected entity. 
Another option is that `+` will be always add at the end, and we could add an option in the `contextmenu` for each entity to `add a child entity` (or whatever)

dmarcos:
Copy / paste could be accessed on:
1. Keyboard shortcut ctrl+c / ctrl + v.
2. Contextual menu or icons associated to the entity.
3. Maybe also next to the `+` icon in the scenegraph.

We can also add a `create child` and `paste` actions to the contextual menu of the entity.

fernandojsg:
Ok, so I'll split all this things into different issues and PR and we could keep discussing them there, as there many little things going here

fernandojsg:
What do we do with this PR? Just merge the way it works right now with the `Edit` menu? and create another PR adding the `clone` and `remove` buttons on the scenegraph`s entity, or modify this PR to include these button too?

dmarcos:
Are you going to include html elment + id in the scenegraph in this PR?

dmarcos:
We can add clone and remove in this PR too I think.

fernandojsg:
> Are you going to include html elment + id in the scenegraph in this PR?

No, I'll leave that for another PR

fernandojsg:
> We can add clone and remove in this PR too I think.

Ok, so I'll add clone and remove buttons next in the scenegraph in this PR and leave the rest for other PRs/issues

fernandojsg:
Btw how do you feel about select the newly created element after clone, or keeping selected the element to be clone?

dmarcos:
I think `+` or clone should select the new element.

fernandojsg:
Updated with the `clone` and `remove` icon added to the scenegraph. When cloning an entity it will select the new entity by default and when removing it will select the closest entity (first looking forward and then backward):
removeclone

ngokevin:
Do we keep the Edit Menu? Don't know if we want more than one way to do something.

src/actions/entity.js
```diff
@@ -38,3 +38,112 @@ export function updateEntity (entity, componentName, propertyName, value) {
...
+  if (entity) {
+    if (force === true || confirm('Do you really want to remove the entity: `' + (entity.id || entity.tagName) + '`')) {
+      var closest = findClosestEntity(entity);
+      entity.parentNode.removeChild(entity);
+      editor.selectEntity(closest);
+    }
+  }
+}
+
+function findClosestEntity(entity) {
```
ngokevin: lint :)

src/actions/entity.js
ngokevin, line 87: yeah, that's fine for now. have the callback pass it

src/components/SceneGraph.js
```diff
@@ -176,8 +178,15 @@ export default class SceneGraph extends React.Component {
...
         return (
           <div key={idx} className={className} value={option.value}
-            dangerouslySetInnerHTML={{__html: option.html}}
-            onClick={() => {this.handleEntityClick(option.value)}}/>
+            onClick={this.setValue.bind(this, option.value)}>
+            <span dangerouslySetInnerHTML={{__html: option.html}}></span>
+              <span className="icons">
+                <a href="#" onClick={() => cloneEntity(option.value)}
+                  title="Clone entity" className="button fa fa-clone"></a>
+                <a href="#" onClick={(e) => {e.stopPropagation(); removeEntity(option.value)}}
```
ngokevin: don't need parens around `e`. can also use `evt` or `event` to be clearer

remove `href`s

src/components/SceneGraph.js
```diff
@@ -176,8 +178,15 @@ export default class SceneGraph extends React.Component {
         let className = 'option' + (option.value === this.state.value ? ' active' : '');
         return (
           <div key={idx} className={className} value={option.value}
-            dangerouslySetInnerHTML={{__html: option.html}}
-            onClick={() => {this.handleEntityClick(option.value)}}/>
+            onClick={this.setValue.bind(this, option.value)}>
```
ngokevin: Can do what you are doing below `() => {this.setValue(option.value);}`

ngokevin:
r+wc

src/actions/entity.js
fernandojsg: 👼

src/components/SceneGraph.js
fernandojsg: done

src/components/SceneGraph.js
fernandojsg: done! I've moved also the tracking's `ga` code from `handleEntityClick` to `setValue`

fernandojsg:
@ngokevin re leaving the options in the `edit` menu, let's leave them there until we'll decide which icons we'll show on the scenegraph. As one of the next things that I'll do is to copy/paste, probably we don't want to have copy, paste, clone and delete in the scenegraph, so we could arrange it in another PR/issue

ngokevin:
The same thing still applies to copy/paste though, you want to keep the direct mapping between the entity and its actions.

fernandojsg:
Yep, but what if we have like 10 actions for the entity, we should leave some of them in the context menu and just the two/three more common on the icons next to the entity data on the scenegraph right?

ngokevin:
Maybe, maybe not. But that's assuming we have 10 actions, which we don't at the moment.

PR #198, opened 2016-07-12
Export scene name based on ID or slugify url. Fixes #22

### LINKED ISSUE #22: Consider generating better filenames than ascene.html for exported scenes

fernandojsg:
Instead of use the current `ascene.html` filename, the editor will take the `id` of `a-scene` if exist or it will slugifly `window.location.host + window.location.pathname` as @cvan proposed on https://github.com/aframevr/aframe-editor/issues/22

### DIFF
--- a/src/components/menu/Menu.js
+++ b/src/components/menu/Menu.js
@@ -1,7 +1,7 @@
-var React = require('react');
-var Clipboard = require('clipboard');
-var Exporter = require('../../lib/exporter.js');
-var Events = require('../../lib/Events.js');
+import React from 'react';
+import Clipboard from 'clipboard';
+import Events from '../../lib/Events.js';
+import {getSceneName, generateHtml} from '../../lib/exporter';
 
 var primitivesDefinitions = {
   'Entity': {group: 'entities', element: 'a-entity', components: {}},
@@ -83,7 +83,7 @@ export class MenuWidget extends React.Component {
   componentDidMount() {
     var clipboard = new Clipboard('[data-action="copy-to-clipboard"]', {
       text: function (trigger) {
-        return Exporter.generateHtml();
+        return generateHtml();
       }
     });
     clipboard.on('error', function(e) {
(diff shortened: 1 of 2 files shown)


src/actions/scene.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Get scene name
+ * @param  {Eelement} scene Scene element
```
ngokevin: typo

src/actions/scene.js
```diff
@@ -0,0 +1,23 @@
...
+  return scene.id || slugify(window.location.host + window.location.pathname);
+}
+
+/**
+ * Slugify the string removing non-word chars and spaces
+ * @param  {string} text String to slugify
+ * @return {string}      Slugified string
+ */
+function slugify(text)
+{
```
ngokevin: brace on same line as function

you can lint this file

src/actions/scene.js
```diff
@@ -0,0 +1,23 @@
+/**
```
ngokevin: these utils can be part of Exporter.js

src/components/menu/Menu.js
```diff
@@ -113,7 +114,8 @@ export class MenuWidget extends React.Component {
       save(new Blob([ text ], { type: 'text/plain' }), filename);
     }
 
-    saveString(Exporter.generateHtml(), 'ascene.html');
+    var sceneName = getSceneName(document.querySelector('a-scene'));
```
ngokevin, line 116: Do we have a more solid reference to the scene? In case of multiple scenes.

src/actions/scene.js
ngokevin: Generally have done docstrings like:

`@param {Element} scene - Scene element.` with only one space, a hyphen to separate argument name and desc, and end with period

ngokevin:
r+wc

src/actions/scene.js
fernandojsg: fixed

src/actions/scene.js
fernandojsg: yep, done!

src/actions/scene.js
fernandojsg: Yep, good idea. Moved!

src/components/menu/Menu.js
fernandojsg, line 116: This is something I've in the TODO, I'll open an issue to see how we could manage multiple scenes in the same page correctly.

PR #201, opened 2016-07-13
Fix #139 adding empty string for null values in inputs

### LINKED ISSUE #139: Console error upon editing components: "Warning: `value` prop on `input` should not be null. Consider using the empty string to clear the component or `undefined` for uncontrolled components."
1. Load the example scene in editor mode.
2. Select an entity in the scene graph.
3. Uncheck the "visible" field.
4. Notice this error in the Console: `Warning:`value`prop on`input`should not be null. Consider using the empty string to clear the component or`undefined`for uncontrolled components.`

image

fernandojsg:


fernandojsg:
So simple, I'll just merge it myself 👼

PR #203, opened 2016-07-13
display tag name brackets in scenegraph

ngokevin:
screen shot 2016-07-13 at 12 22 33 pm

### DIFF
--- a/src/components/SceneGraph.js
+++ b/src/components/SceneGraph.js
@@ -78,7 +78,7 @@ export default class SceneGraph extends React.Component {
   }
 
   rebuildOptions = () => {
-    const options = [{static: true, value: this.props.scene, html: 'a-scene'}];
+    const options = [{static: true, value: this.props.scene, html: '&lt;a-scene&gt;'}];
 
     function treeIterate (element, depth) {
       if (!element) { return; }
@@ -103,7 +103,7 @@ export default class SceneGraph extends React.Component {
           }
 
           const pad = '&nbsp;&nbsp;&nbsp;'.repeat(depth);
-          const label = child.id ? child.id : child.tagName.toLowerCase();
+          const label = child.id ? child.id : '&lt;' + child.tagName.toLowerCase() + '&gt;';
 
           options.push({
             static: true,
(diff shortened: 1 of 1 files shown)


fernandojsg:
cool change!

dmarcos:
Should we do this for all of them?

fernandojsg:
@dmarcos we're displaying `ID || tagName` so this is just for the ones with just `tagName`

dmarcos:
I think it's more consistent and useful if you can see both all the time `<a-entity id="redBox">`. It reinforces the idea that the scene graph is actually the DOM

fernandojsg:
but maybe that will make the scenegraph too wide, just thinking that we need to add icons there for actions for the entities, the visibility toggle icons (eyes),... it could take half of the screen

dmarcos:
I know that adding `<a-scene>` makes the problem a bit worse but the scenegraph being too wide is already a problem with deep hierarchies. How are we going to deal with it?

dmarcos:
Can we find a way to display the html tags without taking too much space? I think it's a good idea.

fernandojsg:
What about something like this? (coder-colors). Adding the `< >` will helps also to differentiate the components' icons (if at the end we'll leave them there) from the entity actions icons, and it will give a better vision of the DOM.
screenshot 2016-07-14 00 28 16

dmarcos:
I really like it. It conveys that those icons are part of the DOM too :). 
If you think about it there won't probably be many elements with IDs in the average experience so the wide scenegraph is less of a problem in practice. People use IDs to mark key elements in the page. I expect to see most of the entities with no ID.

dmarcos:
How do we deal anyways with scengraphs with 3 or more depth levels? Is the scene graph panel adapting already the width?

fernandojsg:
> I expect to see most of the entities with no ID.

Yep you're right

> How do we deal anyways with scenegraphs with 3 or more depth levels? Is the scene graph panel adapting already the width?

Right now is not adapting, but we could make it adapt. I also don't expect to have so many depth levels and even if you have them. 
Another option is to make it fixed size and add scroll when needed (maybe after 3 levels) and the user could horizontal scroll (I know that sucks) or change the size of the sidebar. But probably this one is harder to implement too.

fernandojsg:
Unity even doesn't support horizontal scrolling, you need to resize the scenegraph.
screenshot 2016-07-14 01 18 43

PR #208, opened 2016-07-15
Fix lint errors

fernandojsg:
It fixes most of the lint errors. 
Just a few left:
- @ngokevin the `'ga' is not defined` in multiple files, should I just add `/* global ga */` for each file or we should ignore it in any file from the config?
- `dangerouslySetInnerHTML` will be changed in a future PR to fix some of the scenegraph's code.

### DIFF
--- a/src/lib/editor.js
+++ b/src/lib/editor.js
@@ -35,7 +35,7 @@ Editor.prototype = {
 
     this.editorCameraEl = document.createElement('a-entity');
     this.editorCameraEl.isEditor = true;
-    this.editorCameraEl.addEventListener('loaded', function(entity) {
+    this.editorCameraEl.addEventListener('loaded', function (entity) {
       this.DEFAULT_CAMERA = this.editorCameraEl.getObject3D('camera');
       this.initUI();
     }.bind(this));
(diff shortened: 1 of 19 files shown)


ngokevin:
I'd ignore the `ga` errors for now, I will most likely introduce the cross-origin analytics where I have to add a central helper. So I won't be using the global.

src/lib/editor.js
```diff
@@ -35,7 +35,7 @@ Editor.prototype = {
 
     this.editorCameraEl = document.createElement('a-entity');
     this.editorCameraEl.isEditor = true;
-    this.editorCameraEl.addEventListener('loaded', function(entity) {
+    this.editorCameraEl.addEventListener('loaded', function (entity) {
       this.DEFAULT_CAMERA = this.editorCameraEl.getObject3D('camera');
       this.initUI();
     }.bind(this));
```
ngokevin, line 41: Could also get rid of the `bind(this)` and use `() => {}`

ngokevin:
r+wc/rebase

src/lib/editor.js
fernandojsg, line 41: Yep, the editor.js is a legacy code in es5 used with three.js editor. I'll clean and convert to es6 in another PR

PR #211, opened 2016-07-15
Add link to components' docs. Fixes #31

### LINKED ISSUE #31: Consider adding links to the docs for each component + property
image

Definitely could come later, but it'd be nice if there was some (?) help icon or something that could link me to the docs to learn about it.

fernandojsg:
Added `help` link button for each core component.
screenshot 2016-07-15 19 59 34

### DIFF
--- a/src/components/components/Component.js
+++ b/src/components/components/Component.js
@@ -86,11 +86,19 @@ export default class Component extends React.Component {
       componentName = componentName.substr(0, componentName.indexOf('__'));
     }
 
+    const link = getComponentDocsLink(componentName.toLowerCase());
+    let componentHelp = '';
+    if (link) {
+      componentHelp = <a title='Help' className='button fa fa-question-circle'
+        target='_blank' onClick={event => event.stopPropagation()}
+        href={link}></a>;
+    }
+
     return (
       <Collapsible>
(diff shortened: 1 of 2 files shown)


dmarcos:
Should we alphabetize the labels?

ngokevin:
How about the question mark bubble:

`MATERIAL (?)`

Which would help separate it from the actual actions.

fernandojsg:
yep, I like @ngokevin proposal, it help separate and it's also smaller.

dmarcos:
We could just make the whole component title clickable. The `?` symbol would just reinforce that there's help available.

fernandojsg:
I would expect to click the title and to toggle visibility of the panel, if it will launch the help, I will need to be very careful to click on the empty space outside the actions and title

ngokevin:
I could see collapse attempts unintentionally opening links.

fernandojsg:
Yep, same here. So maybe just use some clickable icon like http://fontawesome.io/icon/info-circle/ or http://fontawesome.io/icon/question-circle/ next to the title?

fernandojsg:
I've added the icon button next to the title:
screenshot 2016-07-19 13 01 22

PR #218, opened 2016-07-19
Fixes #217 reposition and rename edit button

### LINKED ISSUE #217: Reposition and rename "edit/exit" button
From the discussion in https://github.com/aframevr/aframe-editor/issues/214 we're removing the menubar so we need to move the `edit/exit` button too.
I was thinking about changing the name to `Edit scene/Play scene` and to put the button at the botton centered horizontally.

Something like this:
screenshot 2016-07-19 19 58 05
screenshot 2016-07-19 19 57 59

fernandojsg:
screenshot 2016-07-19 20 39 23

### DIFF
--- a/src/components/Main.js
+++ b/src/components/Main.js
@@ -81,9 +81,11 @@ export default class Main extends React.Component {
           </div>
           <ComponentsSidebar/>
         </div>
-        <a className='toggle-edit' onClick={this.toggleEditor}>
-          {this.state.editorEnabled ? 'Exit' : 'Edit'}
-        </a>
+        <div className='toggle-panel'>
+          <a className='toggle-edit-button' onClick={this.toggleEditor}>
+            {this.state.editorEnabled ? 'Play scene' : 'Edit scene'}
+          </a>
+        </div>
       </div>
     );
   }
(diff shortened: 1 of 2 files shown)


dmarcos:
Is the `edit button` also going to be at the center?

dmarcos:
It doesn't seem to get a bit in the way? I understand you want to be consistent with the button position in edit/play mode. On the other hand we should keep the viewport as unobstructed as possible.

fernandojsg:
If we can't use the top (left or right) corners we should probably use the bottom left corner? we could add some bar with the same width of the scenegraph, but I'm not sure, what do you think?

fernandojsg:
Or what about adding a little bar on the top of the scenegraph, maybe with the `AFRAME` as we have before (if you click it will go to aframe.io or whatever) and in the right the `EDIT` button

dmarcos:
We can add a toolbar at the top of the scenegraph. We could add the `export` ,`play`, `+` (to add an entity) buttons there.

dmarcos:
I see a toolbar on top of the scenegraph to contain the actions that globally affect the scene.

fernandojsg:
I'll close this PR to have just one removing the menu and moving the button.

PR #219, opened 2016-07-20
Set pointer cursor for enter/exit button

darkwing:
Minor annoyance but the text cursor is showing over the enter/exit button.  Pointer makes it more apparent it's meant to be clicked.


ngokevin:
Might be worth seeing that all anchor tags have the cursor pointer. Or if that button is an anchor tag

fernandojsg:
@darkwing thanks for catching! I've added cursor pointer to every anchor now in another coming PR.

PR #220, opened 2016-07-20
Make menu title background same as dropdown upon hover

darkwing:
Ties the menu item more closely with the dropdown: 
menuupdate


dmarcos:
We were discussing about removing the top menu bar all together and relocate the actions closer to the context where they apply. For instance, create would be a `+` icon in the scene graph panel

darkwing:
That sounds cool but this can still apply until that work is complete.

PR #226, opened 2016-07-21
builds

ngokevin:


ngokevin:
builds here

fernandojsg:
Cool!! Just one question, `npm run build` will generate `build/aframe-inspector.js` and `npm run dist` will generate `dist/aframe-inspector.js` and the minified version. But the non-minified version is the same in both cases, why do we need both and not just leave one?

fernandojsg:
Ok, solved on slack! thanks

PR #230, opened 2016-08-01
Fixes error while selecting a-scene in scenegraph

fernandojsg:
Fixes an error while selecting a-scene node in the scenegraph. I also fixed the width of the component's name as it wasn't show correctly the ellipsis on overflow.

### DIFF
--- a/src/components/components/Component.js
+++ b/src/components/components/Component.js
@@ -71,7 +71,7 @@ export default class Component extends React.Component {
     const componentData = this.props.component;
 
     if (isSingleProperty(componentData.schema)) {
-      const componentName = componentName.toLowerCase();
+      const componentName = this.props.name;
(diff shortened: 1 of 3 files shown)


src/components/components/Component.js
```diff
@@ -61,7 +61,7 @@ export default class Component extends React.Component {
     const componentData = this.props.component;
 
     if (isSingleProperty(componentData.schema)) {
-      const componentName = componentName.toLowerCase();
+      const componentName = this.props.name.toLowerCase();
```
dmarcos: Why do we need `toLowerCase`?

fernandojsg: I thought that I needed to ensure that `componentName` is lowercase to access the schema for each component and the user could do `GEOMETRY`and it should still works, isn't it?

PR #232, opened 2016-08-02
Add docs links for common components. Fixes #212

### LINKED ISSUE #212: Add help for common/default components
Re https://github.com/aframevr/aframe-editor/issues/31 I created a PR (https://github.com/aframevr/aframe-editor/pull/211) to add the help link for each core components. In the future we'll need to add that feature to the third party components too.
But right now the default components are shown inside the `common` panel and they don't have link to the doc.

fernandojsg:
Added help icon button next to the common components (position, rotation, scale, visible)
screenshot 2016-08-02 23 45 56

### DIFF
--- a/src/actions/component.js
+++ b/src/actions/component.js
@@ -1,3 +1,5 @@
+import React from 'react';
+
 /**
  * Get the list of modified properties
  * @param  {Element} entity        Entity where the component belongs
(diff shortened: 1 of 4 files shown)


src/actions/component.js
```diff
@@ -32,6 +34,35 @@ export function getClipboardRepresentation (entity, componentName) {
...
+ * @param  {string} componentName Component's name
+ * @return {string}               URL to the documentation
+ */
+export function getComponentDocsUrl (componentName) {
+  if (AFRAME.components[componentName]) {
+    // Returns link from the core components
+    return 'https://aframe.io/docs/' + AFRAME.version + '/components/' +
+      (componentName === 'camera' ? '' : componentName.toLowerCase() + '.html');
+  }
+  return null;
```
ngokevin, line 48: Need to return null?

src/components/components/CommonComponents.js
```diff
@@ -42,8 +42,8 @@ const CommonComponents = props => {
             const schema = AFRAME.components[key].schema;
             return (
               <PropertyRow onChange={updateEntity} key={key} name={key}
-                schema={schema} data={componentData.data} componentname={key}
-                entity={props.entity}/>
+                show_help={true} schema={schema} data={componentData.data}
```
ngokevin, line 45: camelCase?

ngokevin:
The help links could be less pronounced (maybe some opacity).

ngokevin:
r+wc

src/components/components/CommonComponents.js
fernandojsg, line 45: yep right.

fernandojsg:
@ngokevin what do you think about something like this (`opacity: 0.4` and `1` on hover)
screenshot 2016-08-03 00 29 59

ngokevin:
I like it.

src/actions/component.js
fernandojsg, line 48: It could be `null, false, undefined...` something to easy check against that it has a valid url.

ngokevin, line 48: I mean if you get rid of this line, it will return undefined by default, which seems suitable.

fernandojsg, line 48: yep, you're right, done.

fernandojsg:
@ngokevin perfect! thank you :)

PR #233, opened 2016-08-02
Using cross-env to set env variables in a cross platform way

dmarcos:
The current `build` and `start` commands are specific to UNIX shells. I replaced them to make them work cross platform. You can either ping me if you modify them next time or make sure you test on Windows.


fernandojsg:
Thanks! :)

Issue #235, opened 2016-08-03
Previous selectbox active when clicking a helper


fernandojsg:
If you've selected a geometry like an sphere you'll see the blue selection box around it, if after that you click a helper (light, camera, and so on) the previous selection box will remains visible.

fernandojsg:
Fixed on https://github.com/aframevr/aframe-inspector/pull/236

PR #236, opened 2016-08-03
Fixes #209 and #235 Selection box errors

### LINKED ISSUE #209: Hide selected boundingbox when object deleted from DOM inspector

fernandojsg:


src/components/Main.js
```diff
@@ -68,7 +66,7 @@ export default class Main extends React.Component {
   render () {
     var scene = document.querySelector('a-scene');
```
dmarcos: Just a suggestion that might not be relevant due to my lack of knowledge of react. Is it worth caching the `sceneEl` element instead of querying it each time we render?

src/lib/viewport/index.js
```diff
@@ -287,8 +287,10 @@ function Viewport (editor) {
     selectionBox.visible = false;
     transformControls.detach();
     if (object !== null) {
-      selectionBox.update(object);
-      selectionBox.visible = true;
+      if ( object.el.getObject3D('mesh')) {
```
dmarcos: extra space after the opening parenthesis

src/components/Main.js
```diff
@@ -30,9 +30,7 @@ export default class Main extends React.Component {
     // Create an observer to notify the changes in the scene
     var target = document.querySelector('a-scene');
     var observer = new MutationObserver(function (mutations) {
```
dmarcos, line 32: This is me again trying to understand how this works. Are we using MutationObservers to check for components updates. We're not using the event `componentchanged` emitted by entities?

fernandojsg, line 32: I'm using `componentchanged` for components updates but when you delete a node for example from the dom inspector or from the console there's no way to catch that the DOM has been modified, so I check for mutations to refresh the scene, and right now I've a special case when the deleted entity (outside from the editor) is the current selected entity, so I need to update the selected entity to avoid inconsistent state

src/lib/viewport/index.js
fernandojsg: thanks!

src/components/Main.js
dmarcos, line 32: in the entities we received a `detachedCallback`. We could emit an event there.

dmarcos:
It looks good to me. For a different PR: It would be great if we could get rid of the MutationObservers and use aframe events instead.

src/components/Main.js
fernandojsg: Yep you're right, it shouldn't be there, I just fixed it.

src/components/Main.js
fernandojsg, line 32: Yep that would be great!

fernandojsg:
Yes, I agree with you that we should try to get rid of MutationObserver and just use aframe events.

PR #239, opened 2016-08-03
Attaches editor to the AFRAME global variable

dmarcos:


fernandojsg:
It will break all the references to `editor` in the rest of the code.

dmarcos:
I think your code should not use the externally exposed API

dmarcos:
changing the public facing API should not affect the code internally

dmarcos:
I made the editor expose a singleton instead of the constructor. The other modules import it instead of relying in the editor variable of the window object. This way we decouple the internal from the publicly exposed API.

fernandojsg:
superb! thank you :)

PR #240, opened 2016-08-03
Massive s/editor/inspector

fernandojsg:
Rename all the variables, events, attributes and so on.

### DIFF
--- a/src/actions/entity.js
+++ b/src/actions/entity.js
@@ -1,4 +1,4 @@
-var editor = require('../lib/editor.js');
+var INSPECTOR = require('../lib/inspector.js');
 var Events = require('../lib/Events.js');
 
 var components = AFRAME.components;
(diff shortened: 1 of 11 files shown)


dmarcos:
there are conflicts

fernandojsg:
@dmarcos sorry, fixed!

src/actions/entity.js
```diff
@@ -1,4 +1,4 @@
-var editor = require('../lib/editor.js');
+var inspector = require('../lib/inspector.js');
```
ngokevin: If inspector is a singleton, might want to cap it.

`INSPECTOR.selectEntity`

fernandojsg: @ngokevin good point, changed!

ngokevin:
r+ from me

fernandojsg:
Thanks!

PR #248, opened 2016-08-04
Apply Array.slice to use foreach on nodelists

fernandojsg:
To fix compatibility issues on firefox


darkwing:
`Array.prototype.slice.call(qsa, 0)`

fernandojsg:
@darkwing ah ok, I thought that by default it will take 0, changing! thank you :)

fernandojsg:
@darkwing I was just checking this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice and if begin is not defined it will take `0` as default so I believe we don't need to add the extra parameter right?

PR #251, opened 2016-08-08
Fixes #165 Hiding sidebar when no entity is selected

### LINKED ISSUE #165: Consider hiding "Attributes" sidebar when no scene-graph entity is selected
1. Load a scene with the editor mode.
2. Notice an empty "Attributes" sidebar appears:

image

fernandojsg:
hidesidebar

### DIFF
--- a/src/components/components/Sidebar.js
+++ b/src/components/components/Sidebar.js
@@ -51,24 +51,24 @@ export default class Sidebar extends React.Component {
 
   render () {
     const entity = this.state.entity;
-    let entityButtons = '';
-    let entityName = '';
     if (entity) {
-      entityButtons = <div>
+      const entityButtons = <div>
         <a href='#' title='Copy entity HTML to clipboard' data-action='copy-entity-to-clipboard'
           className='button fa fa-clipboard' onClick={event => event.stopPropagation()}></a>
       </div>;
-      entityName = '<' + entity.tagName.toLowerCase() + '>';
-    }
+      const entityName = '<' + entity.tagName.toLowerCase() + '>';
 
-    return (
-      <div id='sidebar'>
-        <div className='sidebar-title'>
-          <code>{entityName}</code>
-          {entityButtons}
+      return (
+        <div id='sidebar'>
+          <div className='sidebar-title'>
+            <code>{entityName}</code>
+            {entityButtons}
+          </div>
+          <ComponentsContainer entity={this.state.entity}/>
         </div>
-        <ComponentsContainer entity={this.state.entity}/>
-      </div>
-    );
+      );
+    } else {
+      return <div/>;
+    }
   }
 }
--- a/src/lib/inspector.js
+++ b/src/lib/inspector.js
@@ -177,16 +177,16 @@ Inspector.prototype = {
       this.createNewEntity(definition);
     }.bind(this));
 
-    Events.on('domModified', (mutations) => {
+    Events.on('domModified', mutations => {
       var self = this;
-      mutations.forEach((mutation) => {
-        if (mutation.type === 'childList') {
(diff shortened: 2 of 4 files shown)


src/components/components/Sidebar.js
```diff
@@ -31,24 +31,24 @@ export default class Sidebar extends React.Component {
...
+            {entityButtons}
+          </div>
+          <ComponentsContainer entity={this.props.entity}/>
         </div>
-        <ComponentsContainer entity={this.props.entity}/>
-      </div>
-    );
+      );
+    } else {
+      return <div></div>
```
ngokevin: Could do `<div/>`

ngokevin:
There's no way to hide the sidebar after initial entity select (e.g., clicking on empty space of scene graph, clicking on scene, close panel option).

fernandojsg:
I'm fixing these issues, but one is already on the way (selecting a-scene) on this PR: https://github.com/aframevr/aframe-inspector/pull/230

fernandojsg:
@ngokevin I've made some changes that should works on all the cases except while clicking on a-scene, could you give it a try?

fernandojsg:
Updated after merging a dependent PR (https://github.com/aframevr/aframe-inspector/pull/230)

fernandojsg:
@ngokevin could you take a look at this one?

src/lib/inspector.js
```diff
@@ -179,15 +179,17 @@ Inspector.prototype = {
...
-        if (mutation.type === 'childList') {
-          Array.prototype.slice.call(mutation.removedNodes).forEach(function (removedNode) {
-            if (self.selectedEntity === removedNode) {
-              self.selectEntity(null);
-            }
-          });
-        }
-      });
+      if (mutations) {
+        mutations.forEach((mutation) => {
```
ngokevin: Don't need parens if single-arg

src/lib/inspector.js
```diff
@@ -179,15 +179,17 @@ Inspector.prototype = {
...
+          if (mutation.type === 'childList') {
+            Array.prototype.slice.call(mutation.removedNodes).forEach(function (removedNode) {
```
ngokevin: Don't need `self` if you use arrow binding syntax

src/lib/inspector.js
```diff
@@ -179,15 +179,17 @@ Inspector.prototype = {
...
-      mutations.forEach((mutation) => {
-        if (mutation.type === 'childList') {
-          Array.prototype.slice.call(mutation.removedNodes).forEach(function (removedNode) {
-            if (self.selectedEntity === removedNode) {
-              self.selectEntity(null);
-            }
-          });
-        }
-      });
+      if (mutations) {
```
ngokevin: If you return early, you could reduce the indent of the block below

`if (!mutations) { return; }`

`if (mutation.type !=='childList') { return; }`

ngokevin:
r+wc
- Later would be nice to add a transition to the panel hide.
- I assume clicking on scene graph empty space to hide panel is fixed too.

Just to note, here's how MagicaVoxel does it. Don't have to do it this way, I think this pattern only works better if there are multiple panels:

magica

src/lib/inspector.js
fernandojsg: Yep, missed that, thanks!

src/lib/inspector.js
fernandojsg: Yep, fixed!

src/lib/inspector.js
fernandojsg: Ok, changed!

fernandojsg:
Ok ,thank you. I'll file issues with the proposed changes

PR #254, opened 2016-08-10
Fixes #253 Remove document listener for "componentchanged"

### LINKED ISSUE #253: Listen to `componentchanged` event on individual entities rather than on `document`
I'd like to stop bubbling of the `componentchanged` event as it's better if you're developing and theoretically more performant as components are constantly changing. https://github.com/aframevr/aframe/pull/1758

It was reverted since Inspector relies on bubbling. But we can update it so it doesn't:
- SceneGraph uses it to show the icons. Instead have the Component Panel communicate to the SceneGraph when a component is added.
- `lib/editor.js` uses it to new objects. Can we use `child-attached`?
- `lib/viewport.js` - I would also have the Component Panels emit its own event rather than use the A-Frame event.

fernandojsg:


src/components/components/CommonComponents.js
```diff
@@ -21,6 +22,14 @@ export class CommonComponents extends React.Component {
     entity: React.PropTypes.object
   };
 
+  componentDidMount () {
+    Events.on('selectedEntityComponentChanged', detail => {
+      if (DEFAULT_COMPONENTS.indexOf(detail.name) !== -1) {
+        this.forceUpdate();
```
fernandojsg: @ngokevin what do you think about it? Maybe change for the approach of the second answer here http://stackoverflow.com/questions/30626030/can-you-force-a-react-component-to-rerender-without-calling-setstate where it just updates the key generating a random number?

ngokevin: Yeah, if that works. `redux` might help later on if all the component data is in a single store.

fernandojsg: When you say `that works` do you mean the current code or the stackoverflow's version? :D

ngokevin: sorry, the StackOverflow version

fernandojsg: @ngokevin changed! It seems to works perfect, as you change the state it will trigger render again.

src/components/components/CommonComponents.js
```diff
@@ -21,6 +22,21 @@ export class CommonComponents extends React.Component {
     entity: React.PropTypes.object
   };
 
+  constructor (props) {
+    super(props);
+    this.state = {
```
ngokevin, line 27: Might want to comment the instances of the key change refresh right?

ngokevin:
r+ thanks!

src/components/components/CommonComponents.js
fernandojsg, line 27: Done! :)

fernandojsg:
Thanks you!

PR #258, opened 2016-08-11
Fix 'Can only update a mounted or mounting component'

fernandojsg:
With the latest changes trying to prevent `forceUpdate()` on `componentchanged` I generated a random key for the component, but this is generating the component to regenerate itself and some update events from the widgets are conflicting with the mounting of the component as is stated on the console:

```
warning.js:44 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
```

I've changed it back to `forceUpdate()` to get rid of this until we'll find a better solution. /cc @ngokevin

### DIFF
--- a/src/components/components/CommonComponents.js
+++ b/src/components/components/CommonComponents.js
@@ -24,17 +24,12 @@ export class CommonComponents extends React.Component {
 
   constructor (props) {
     super(props);
-    this.state = {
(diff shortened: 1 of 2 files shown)


ngokevin:
r+, redux would solve all of this later anyways

fernandojsg:
Perfect! Thanks

Issue #160, opened 2016-07-08
Introduce Roboto Mono font family


cvan:
https://www.google.com/fonts/specimen/Roboto+Mono

and add "Roboto Mono" to our mono `font-family`: https://github.com/aframevr/aframe-editor/blob/f408ba7/src/css/dark.css#L774-L776

cvan:
Also available for download from https://github.com/google/fonts/archive/master.zip

Coder206:
@cvan I would like to work on this.

fernandojsg:
@Coder206 cool. Please note that there're other two issues related to this one, if you could work on them it will be great too:
https://github.com/aframevr/aframe-inspector/issues/136
https://github.com/aframevr/aframe-inspector/issues/132

Coder206:
@fernandojsg Absolutely! :)

Coder206:
I don't see dark.css in the master branch. How do I proceed to make a PR?

fernandojsg:
@Coder206 dark.css styles were inserted on main.css, so we've just one CSS right now. Feel free to check it there

Coder206:
@fernandojsg Thanks!

cvan:
Fixed by @Coder206 in PR #272 😄

Issue #136, opened 2016-07-07
Gracefully degrade (and use JS) to inject font-awesome and Google Fonts CSS


cvan:
If maxcdn/Google Fonts takes a while or if I'm offline, the Editor will take a long time/forever to load.

![image](https://cloud.githubusercontent.com/assets/203725/16649060/df468a2e-43eb-11e6-97ac-26e50dca5d9c.png)

[This](https://github.com/MozVR/iswebvrready/blob/7974601/src/_base.html#L12) is one approach (using https://github.com/typekit/webfontloader). Also, there's now a [Font Loading API](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization#optimizing-font-rendering-with-the-font-loading-api) that ships with Chrome.

fernandojsg:
@cvan this was fixed on https://github.com/aframevr/aframe-inspector/pull/272 right?

fernandojsg:
Fixed by https://github.com/aframevr/aframe-inspector/pull/272

cvan:
Now, the `.css` file isn't a blocking request. But, still, with how the code's set up right now, the `.css` file plus all the font files (e.g., `.woff2) still have to still fetch.

Because of the two issues mentioned below, the browser renders all the text as blank until all the requests finish. FWIW, modern browsers recently added a new Font Loading API (but the [Web Font Loader](https://github.com/typekit/webfontloader) JS prolyfill was already added in PR #272, which works fine), but it doesn't handle Google Fonts well (because of the various character sets served, varied based on the HTTP request headers).

Anyway, when the Inspector gets injected, wherever `Roboto` and `Robot Mono` (which is everywhere), the text is invisible until all the font assets load, at which point the text is then rendered using the loaded respective web fonts.

* When the `WebFont` loader was added (in PR #272), it was added for [`Roboto Mono`](https://github.com/aframevr/aframe-inspector/blob/521a0ec/src/components/Main.js#L131), but not for [`Roboto` (sans-serif)](https://github.com/aframevr/aframe-inspector/blob/521a0ec/src/components/Main.js#L22) (and `Roboto` is used in [`main.css`](https://github.com/aframevr/aframe-inspector/blob/521a0ec/src/css/main.css) [here](https://github.com/aframevr/aframe-inspector/blob/521a0ec/src/css/main.css#L3) and [here](https://github.com/aframevr/aframe-inspector/blob/521a0ec/src/css/main.css#L3)).
[...]

Issue #132, opened 2016-07-07
Roboto `font-family` not used for default `input, select, textarea`


cvan:
#### before

![image](https://cloud.githubusercontent.com/assets/203725/16648596/bc0d4392-43e9-11e6-812b-0ab18d617f13.png)
#### after `input, textarea, select { font-family: inherit }`

![image](https://cloud.githubusercontent.com/assets/203725/16648614/ca767bb0-43e9-11e6-8add-3e14fe7c567e.png)

cvan:
Fixed by @Coder206 in PR #272 😄

PR #272, opened 2016-08-13
Introduction of Roboto Mono

Coder206:
This addresses #132, #136 and #160

### DIFF
--- a/src/components/Main.js
+++ b/src/components/Main.js
@@ -96,10 +96,17 @@ function injectCSS (url) {
   link.rel = 'stylesheet';
   link.media = 'screen,print';
   link.setAttribute('data-aframe-inspector', 'style');
-  document.getElementsByTagName('head')[0].appendChild(link);
+  document.head.appendChild(link);
 }
 
 (function init () {
+  var webFont = document.createElement('script');
+  webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
+  var webFontLoader = document.createElement('script');
+  webFontLoader.innerHTML = 'WebFont.load({google: {families: ["Roboto Mono"]}});';
+  document.head.appendChild(webFont);
+  document.head.appendChild(webFontLoader);
+
   var div = document.createElement('div');
   div.id = 'aframe-inspector';
   div.setAttribute('data-aframe-inspector', 'app');
--- a/src/components/Main.js
+++ b/src/components/Main.js
@@ -100,12 +100,18 @@ function injectCSS (url) {
 }
 
 (function init () {
-  var webFont = document.createElement('script');
-  webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
   var webFontLoader = document.createElement('script');
   webFontLoader.innerHTML = 'WebFont.load({google: {families: ["Roboto Mono"]}});';
+
+  var webFont = document.createElement('script');
+  webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
+  webFont.addEventListener('load', function () {
+    document.head.appendChild(webFontLoader);  
+  });
+  webFont.addEventListener('error', function () {
+    console.warn('Could not load WebFont script:', webFont.src);
+  });
   document.head.appendChild(webFont);
-  document.head.appendChild(webFontLoader);
 
   var div = document.createElement('div');
   div.id = 'aframe-inspector';
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -587,8 +587,20 @@ span.subcomponent {
   cursor: grabbing;
 }
 
-code, pre {
+code,
+pre,
+input,
+textarea,
+select {
   font-family: Consolas, Andale Mono, Monaco, Courier New, monospace;
+}
+
+.wf-active code,
+.wf-active pre,
+.wf-active input,
+.wf-active textarea,
+.wf-active select {
+  font-family: Roboto Mono, Consolas, Andale Mono, Monaco, Courier New, monospace;
 }
 
 .tagName {


src/components/Main.js
```diff
@@ -87,17 +83,13 @@ export default class Main extends React.Component {
   }
 }
 
-function injectCSS (url) {
-  var link = document.createElement('link');
-  link.href = url;
-  link.type = 'text/css';
-  link.rel = 'stylesheet';
-  link.media = 'screen,print';
-  link.setAttribute('data-aframe-editor', 'style');
```
fernandojsg: It's important to keep adding `data-aframe-editor="style"` attribute to the injected CSS to remove it while exporting the scene

Coder206: @fernandojsg So do I need to keep the injectCSS?

fernandojsg: I've never used webfont, but maybe you can inject somehow the attributes on the code that it will generate?

src/components/Main.js
```diff
@@ -87,17 +83,13 @@ export default class Main extends React.Component {
...
-  document.getElementsByTagName('head')[0].appendChild(link);
-}
-
 (function init () {
+  var webFont = document.createElement('script');
```
Coder206, line 106: @fernandojsg So in this code, you would want me to inject a data attribute?

src/components/Main.js
```diff
@@ -87,17 +83,13 @@ export default class Main extends React.Component {
...
+  webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
+
+  var webFontLoader = document.createElement('script');
+  webFontLoader.innerHTML = "WebFont.load({google: {families: ['Roboto Mono']}});";
```
fernandojsg: Where is included the `WebFont` dependency? I don't see it in the code

Coder206: @fernandojsg I don't think it is in the code, I thought line 88 was the dependency. Line 88 is referencing the webfont.js file.

fernandojsg: Ok, I think the problem is that you've removed also the `font-awesome` font, which is needed for the icons in the scenegraph, I was trying your branch and I couldn't get the icons on it so I thought the dependency was failing

src/components/Main.js
```diff
@@ -13,10 +13,6 @@ import ToolBar from './ToolBar';
 
 import '../css/main.css';
 
-// Megahack to include font-awesome.
-injectCSS('https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css');
```
fernandojsg, line 16: Apart from `Roboto`, `font-awesome` is also needed for the icons

src/components/Main.js
Coder206: @fernandojsg That makes sense. Sorry about that. 

Was `font-awesome` loaded using injectCSS()?

src/components/Main.js
Coder206, line 16: @fernandojsg I think I can add `font-awesome` and `Roboto` (as already done) using web font.

fernandojsg, line 16: Yep I think so

Coder206:
@fernandojsg Instead of using JavaScript, perhaps we could use CSS - it looks a lot simpler to implement? (https://hacks.mozilla.org/2016/06/webfont-preloading-for-html5-games/)
What are your thoughts on this?

fernandojsg:
I don't have an opinion about what could be better, maybe @cvan that opened the issues could help here

Coder206:
@fernandojsg @cvan Could I get a status update? (I would like to know the game plan before rebasing)

src/css/main.css
```diff
@@ -870,7 +870,7 @@ span.subcomponent {
 }
 
 code, pre {
-  font-family: Consolas, Andale Mono, Monaco, Courier New, monospace;
+  font-family: 'Roboto Mono' Consolas, Andale Mono, Monaco, Courier New, monospace;
```
cvan: forgot a comma after `'Roboto Mono'`

src/css/main.css
```diff
@@ -991,3 +991,7 @@ a.flat-button:hover {
 	display: flex;
 	align-items: center;
 }
+
+input, textarea, select {
```
cvan: why not combine this with the above lines?

src/css/main.css
cvan: you should listing `'Roboto Mono'` only when WebFont has loaded, like this:

``` css
.wf-active code, .wf-active pre {
```

and then keep the others set.

src/css/main.css
cvan: and same as above.

cvan:
could you include the commit message in your commit?

Coder206:
@cvan What do you mean by commit message? (Doesn't doing `git commit -m "Message"` qualify as one?)

cvan:
sorry, I meant `Combining CSS queries (fixes #272)`. but it's not a huge deal 😄

src/components/Main.js
```diff
@@ -99,7 +98,11 @@ function injectCSS (url) {
   document.getElementsByTagName('head')[0].appendChild(link);
 }
 
-(function init () {
+(function init() {
+  var webFont = document.createElement('script');
+  webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
+  var webFontLoader = document.createElement('script');
+  webFontLoader.innerHTML = "WebFont.load({google: {families: ['Roboto Mono']}});";
```
cvan: to be consistent, I would change to this:

``` js
  webFontLoader.innerHTML = 'WebFont.load({google: {families: ["Roboto Mono"]}});';
```

src/css/main.css
```diff
@@ -587,8 +587,8 @@ span.subcomponent {
   cursor: grabbing;
 }
 
-code, pre {
-  font-family: Consolas, Andale Mono, Monaco, Courier New, monospace;
+.wf-active code, .wf-active pre, .wf-active input, .wf-active textarea, .wf-active select {
```
cvan: this is correct, but I would suggest putting each selector on a newline to improve readability

src/css/main.css
```diff
@@ -587,8 +587,8 @@ span.subcomponent {
   cursor: grabbing;
 }
 
-code, pre {
```
cvan, line 590: you should actually keep this _and_ also add the following `.wf-active` block

src/components/Main.js
```diff
@@ -14,10 +14,9 @@ import '../css/main.css';
 
 // Megahack to include font-awesome.
 injectCSS('https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css');
-injectCSS('https://fonts.googleapis.com/css?family=Roboto:400,300,500');
 
 export default class Main extends React.Component {
-  constructor (props) {
```
cvan, line 20: did you make these changes, or did your editor? based on Standard/Semistandard rules, this is actually correct to keep a space before

src/components/Main.js
```diff
@@ -99,7 +98,11 @@ function injectCSS (url) {
   document.getElementsByTagName('head')[0].appendChild(link);
 }
 
-(function init () {
+(function init() {
+  var webFont = document.createElement('script');
+  webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
+  var webFontLoader = document.createElement('script');
```
cvan: where do you actually add `webFontLoader` to `document.head`? how is this script getting loaded?

src/components/Main.js
```diff
@@ -89,7 +88,7 @@ export default class Main extends React.Component {
   }
 }
 
-function injectCSS (url) {
+function injectCSS(url) {
```
cvan: it seems odd to have `injectCSS` but not `injectJS`

src/components/Main.js
```diff
@@ -99,7 +98,11 @@ function injectCSS (url) {
   document.getElementsByTagName('head')[0].appendChild(link);
```
cvan: can you change to `document.head`?

cvan:
thanks for working on this! it's very close 👍

cvan (changes requested):

Coder206:
@cvan Thanks the feedback! Looking to competing your requests. I would like to chat with you more about this PR on slack

cvan:
@Coder206 I didn't understand your question on Slack. could you comment beneath the lines or comments you were unclear on? thanks - and sorry for the delays!

src/components/Main.js
Coder206: @cvan This was editor I suppose, I was working on another section.

src/components/Main.js
Coder206: @cvan Indeed, did you want me to add it?

src/css/main.css
Coder206: @cvan So have just `code, pre, input... { font-family: 'Roboto Mono', ...}` and this too: `.wf-active { font-family: 'Roboto Mono', ... }`?

src/components/Main.js
Coder206: Sorry about that! I am going to add it.

Coder206:
@cvan It's fine! I made some comments.

src/components/Main.js
cvan: if we're going to ever inject another JS file into the page, yes. [the `injectCSS` calls at the top](https://github.com/aframevr/aframe-inspector/pull/272/files#diff-ea337f34aa79e31b41de2123664a0de4R16) just struck me as odd. it's fine to keep for now.

src/components/Main.js
cvan: let me know when you've added that line

src/css/main.css
cvan: no, you don't want `Roboto Mono` in both instances. because, the whole point of the Google Web Font Loader is to allow you to not have the white Flash Of Invisible Text ([FOIT](https://css-tricks.com/fout-foit-foft/)). but you have to use the classes (e.g., `wf-active`) that are set by the Web Font Loader that tell you when `Roboto Mono` has loaded. otherwise, the browser will wait for `Roboto Mono` to load before rendering _any_ text. so the approach would be to have all these lines:

``` css
code,
pre,
input,
textarea,
select {
  font-family: Consolas, Andale Mono, Monaco, Courier New, monospace;
}

.wf-active code,
.wf-active pre,
.wf-active input,
.wf-active textarea,
.wf-active select {
  font-family: Roboto Mono, Consolas, Andale Mono, Monaco, Courier New, monospace;
}
```

hope that makes sense.

cvan (changes requested):

Coder206:
@cvan This makes sense!

Coder206:
@cvan The changes you requested have been made :)

src/components/Main.js
```diff
@@ -96,10 +96,17 @@ function injectCSS (url) {
...
+  document.head.appendChild(link);
 }
 
 (function init () {
+  var webFont = document.createElement('script');
+  webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
+  var webFontLoader = document.createElement('script');
+  webFontLoader.innerHTML = 'WebFont.load({google: {families: ["Roboto Mono"]}});';
+  document.head.appendChild(webFont);
+  document.head.appendChild(webFontLoader);
```
cvan: Using `document.createElement` + `appendChild` to create and inject a `<script>` together result in the creation, injection, and loading of a script asynchronously.

I haven't tested this pull request locally, but I'm almost positive that you have to wait for the `webFont` script to be created and injected before you inject the `webFontLoader` script.

You should make the following changes, and things should work just fine:

``` js
var webFont = document.createElement('script');
webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
var webFontLoader = document.createElement('script');
webFontLoader.innerHTML = 'WebFont.load({google: {families: ["Roboto Mono"]}});';
document.head.appendChild(webFont);
document.head.appendChild(webFontLoader);
```

to

``` js
var webFontLoader = document.createElement('script');
webFontLoader.innerHTML = 'WebFont.load({google: {families: ["Roboto Mono"]}});';

var webFont = document.createElement('script');
webFont.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
webFont.addEventListener('load', function () {
  document.head.appendChild(webFontLoader);  
});
webFont.addEventListener('error', function () {
  console.warn('Could not load WebFont script:', webFont.src);
});
document.head.appendChild(webFont);
```

src/css/main.css
```diff
@@ -587,10 +587,22 @@ span.subcomponent {
...
+select {
   font-family: Consolas, Andale Mono, Monaco, Courier New, monospace;
 }
 
+.wf-active code,
+.wf-active pre,
+.wf-active input,
+.wf-active textarea,
+.wf-active select {
+  font-family: Roboto Mono, Consolas, Andale Mono, Monaco, Courier New, monospace;
```
cvan, line 603: perfect 👍

cvan (changes requested):
thanks, @Coder206! changes look good.

sorry, I missed one thing. so just one more last change. my bad for not catching it earlier - totally my fault.

Coder206:
@cvan Done :) (No worries) Works really well on my end. (I took me a while to find the A-Frame code inside the sockjs server code)

cvan:
r+ this looks perfect - many thanks for your contribution and patience!

PR #280, opened 2016-08-16
Fixes #274 prevent overload the default scene camera

### LINKED ISSUE #274: Camera position is not honored
Open / Close the inspector with the composite example (no entities modified). This is before opening:

Before

This is after closing the inspector:

After

fernandojsg:
Remove the `default` camera attribute and set `isInspector` tag, so it won't be exported while saving the scene.

### DIFF
--- a/src/lib/inspector.js
+++ b/src/lib/inspector.js
@@ -33,10 +33,17 @@ Inspector.prototype = {
     this.container = document.querySelector('.a-canvas');
     this.currentCameraEl = document.querySelector('[camera]');
 
+    // If the current camera is the default, we should prevent AFRAME from
+    // remove it once when we inject the editor's camera
+    if (this.currentCameraEl.hasAttribute('data-aframe-default-camera')) {
+      this.currentCameraEl.removeAttribute('data-aframe-default-camera');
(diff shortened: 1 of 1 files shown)


src/lib/inspector.js
```diff
@@ -33,22 +33,29 @@ Inspector.prototype = {
     this.container = document.querySelector('.a-canvas');
     this.currentCameraEl = document.querySelector('[camera]');
 
+    // If the current camera is the default, we should prevent AFRAME from
+    // remove it once when we inject the editor's camera
+    if (this.currentCameraEl.hasAttribute('data-aframe-default-camera')) {
+      this.currentCameraEl.removeAttribute('data-aframe-default-camera');
+      this.currentCameraEl.isInspector = true;
```
dmarcos: We could add an attribute to be consistent with a-frame. Something like `el.setAttribute('inspector-injected')`?

fernandojsg: Yes, I forgot to add also the injected attribute, the `isInspector` is to be used in the same way as `isEntity`

fernandojsg: In fact I didn't need `isInspector` on that, just the attribute :) good catch! fixed

PR #283, opened 2016-08-16
Add components icons title with parameters and hover color

fernandojsg:
Added style (color) when mouse is over the component's icons, and added a title to the icon with the component's name and its properties values (non-default ones) for quick review.
screenshot 2016-08-17 00 33 37


ngokevin:
nice I'd sort the props r+wc

fernandojsg:
@ngokevin yep, good idea. Done!
screenshot 2016-08-17 00 51 10

PR #287, opened 2016-08-17
Remove unused CSS and convert function expressions to arrow functions

fernandojsg:


src/lib/inspector.js
```diff
@@ -42,19 +45,17 @@ Inspector.prototype = {
 
     this.inspectorCameraEl = document.createElement('a-entity');
     this.inspectorCameraEl.isInspector = true;
-    this.inspectorCameraEl.addEventListener('loaded', function (entity) {
+    this.inspectorCameraEl.addEventListener('loaded', entity => {
```
dmarcos, line 48: I actually prefer named functions because the name will show up in the stack trace when using the debugger.

``` javascript
  this.inspectorCameraEl.addEventListener('loaded', function onLoaded(entity) { ... });
```

In general I don't think arrow functions add any value (many will be probably shocked). It has implicit magic (binding) and makes syntax less tight and redundant because they introduce multiple ways to express one thing (a function expression)

fernandojsg, line 48: The rest of the code is already using arrow functions as it's the recommended for es6&react we just had this legacy code based on threejs' editor, so If we want to go back to named function we should discuss it and create another issue/PR to rewrite the rest of the code.

PR #292, opened 2016-08-17
update README image using different scene

ngokevin:
Let me know if you think this screen-cap looks better:

screen shot 2016-08-17 at 2 47 45 pm


fernandojsg:
supercool thanks!

PR #293, opened 2016-08-17
Fix #288 Remove helpers when object is deleted

### LINKED ISSUE #288: Remove helpers when the parent object is removed

fernandojsg:


src/lib/inspector.js
```diff
@@ -97,22 +97,14 @@ Inspector.prototype = {
...
-  removeObject: function (object) {
-    if (object.parent === null) return; // avoid deleting the camera or scene
-
-    var scope = this;
 
-    object.traverse(function (child) {
-      scope.removeHelper(child);
+  removeObject: function (object) {
+    // Remove just the helper as the object will be deleted by Aframe
+    object.traverse(child => {
```
ngokevin: Would `object.traverse(this.removeHelper)` work?

ngokevin:
r+

src/lib/inspector.js
fernandojsg: Yep, I was refactoring that function and didn't realised that it was just the function call hehe. 
Actually I believe it should be: `object.traverse(this.removeHelper).bind(this)`

PR #295, opened 2016-08-18
Fixes #275 Working with embedded scenes

### LINKED ISSUE #275: The inspector does not work with embedded scenes
Load the wikipedia example. This is what I see:

wilkipedia

fernandojsg:
With the help of some code stolen from https://github.com/aframevr/aframe-inspector/pull/294 by  @ngokevin ❤️ 

wikipedia

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -2,6 +2,9 @@ body {
   font-family: Helvetica, Arial, sans-serif;
   font-size: 14px;
   margin: 0;
+}
+
+body.aframe-inspector-opened {
   overflow: hidden;
 }
 
@@ -162,7 +165,7 @@ input.number:focus {
 }
 
 #left-sidebar {
-  position: absolute;
+  position: fixed;
   left: 0;
   top: 0;
   bottom: 0;
@@ -727,6 +730,10 @@ a.help-link:hover {
   display: flex;
   justify-content: flex-end;
   align-items: stretch;
+  position: fixed;
+  top: 0;
+  bottom: 0;
+  right: 0;
 }
 
 #aframe-inspector-panels {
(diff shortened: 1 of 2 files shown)


src/css/main.css
```diff
@@ -736,6 +743,10 @@ a.help-link:hover {
 
 /* This is a temporaly hack, we should style the editor instead of overwriting
    the a-scene to fix the "display: block" issue. */
-a-scene {
+.aframe-inspector-opened a-scene {
```
ngokevin, line 751: What about multiple scenes? Maybe we should keep track of the actively inspected scene.

ngokevin:
Got some peeking DOM elements. What if we just `display: none !important` everything that is not the canvas?

screen shot 2016-08-17 at 6 02 28 pm

src/css/main.css
fernandojsg, line 751: We should keep track of that yep, but right now there's no support for multiple scenes on the rest of the code as we're looking for example for `a-scene` in query selector. So we should do another PR to achieve that, and also expose the other scenes on the inspector UI probably so you could switch them.

ngokevin:
Would be good to hide every element on the page:

screen shot 2016-08-17 at 6 03 16 pm

src/css/main.css
ngokevin, line 751: ok np

fernandojsg:
@ngokevin umm what about a background-color for `canvas` ?

fernandojsg:
and `z-index` for canvas and UI ?

ngokevin:
Yeah, could work if we save/restore the background color and add a high-enough z-index.

fernandojsg:
I've added `background-color` to `canvas` and `z-index` to `canvas` and `panels` but I can't get rid of some elements in the top of the canvas :| Any idea @ngokevin ?
screenshot 2016-08-18 03 26 23
screenshot 2016-08-18 03 26 14

ngokevin:
r+

Issue #298, opened 2016-08-18
Suggestions for new shortcuts


feiss:
Following #154 , support for more shortcuts would be very interesting for desktop users. 

(Some of these actions are not yet included in the editor, I'll try to write a spec for each one in a separated issue)
- [x]  **`w,e,r`**
  Switch to Translate/Rotate/Scale modes (Maya and Max standard, **very convenient**)
- [x] **`z`** 
  View selected object (**f** would be a more convenient shortcut for this, but is currently used for 'fullscreen') (see #300)
- [ ] **`a`** 
  View all objects (see #300 )
- [ ] **`ctrl + LMB`** 
  Add object to selection (this could be an whole issue itself, support for multiple selection)
- [ ] **`ctrl + a`**
  Select All
- [x] **`d`**
  Duplicate/clone current selection
- [x] **`g`** 
  Toggle grid visibility
- [x] **`n`**
  Add new entity
- [x]  **`supr/backspace`** 
  Delete current selection
- [x] **`ctrl+c, ctrl+v`**
  copy/paste current selection
- [x] **`1,2,3,4`**
  Switch to Front/Left/Top/Perspective view
- [x] **`o`**
  Toggle local/global transform mode
- [ ] **`Num +, Num -`** 
  Increase/decrease transform helper (in discrete steps, between (min, max) values)
- [x] **`ctrl+alt+i`** 
  Switch Edit and VR modes. 
- [x] **`Esc`** for exiting edit mode (current behaviour)
- [x] **`h`** 
  Toggle help/shortcuts panel

Shortcut customization in the cheatsheet panel that #276 suggested would be nice, although I don't think it's high priority.

fernandojsg:
Thanks! Regarding the `esc` I was thinking the same but @dmarcos pointed that maybe the VR-scene is capturing `esc` by itself as it's a very generic shortcut so we can't just hijack it :(

ngokevin:
When the inspector is active, the scene should be paused and event listeners should be removed, so keypresses should not do anything.

fernandojsg:
@ngokevin yes that's why `esc` will works while the editor is opened, but if we exit the inspector and go back to the scene we should remove that listener

feiss:
Can't you just add the event listener after the vr-scene and its events are initialized? (maybe with a setTimeout()?)

fernandojsg:
[...]

PR #305, opened 2016-08-18
Add basic shortcuts

fernandojsg:
Based on https://github.com/aframevr/aframe-inspector/issues/298 I added some shortcuts:
- **`w,e,r`**
  Switch to Translate/Rotate/Scale modes
- **`d`**
  Duplicate/clone current selection
- **`g`** 
  Toggle grid visibility
- **`n`**
  Add new entity
- **`supr / backspace`** 
  Delete current selection
- **`ctrl+alt+i`** 
  Switch Edit and VR modes. 
- **`Esc`** for exiting edit mode (current behaviour)

shortcuts

### DIFF
--- a/src/actions/entity.js
+++ b/src/actions/entity.js
@@ -1,4 +1,3 @@
-var INSPECTOR = require('../lib/inspector.js');
 var Events = require('../lib/Events.js');
 var components = AFRAME.components;
 var isSingleProperty = AFRAME.schema.isSingleProperty;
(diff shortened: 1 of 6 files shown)


src/actions/entity.js
```diff
@@ -1,4 +1,3 @@
-var INSPECTOR = require('../lib/inspector.js');
```
fernandojsg, line 1: @dmarcos Don't know why but accessing to `INSPECTOR` returns empty object instead of the actual inspector, so I just changed everything here to use `AFRAME.INSPECTOR` but it should works that way. Any idea what could be happening?

ngokevin:
might want something different than `d` (delete?) for duplicate

fernandojsg:
@ngokevin umm I thought about `c` (clone) but don't know if we'll use that for `copy`. I believe people will try first to use `backspace` or `supr` to delete so we won't need another shortcut for that, in any case we could also use `r` (remove) as it's the way we name it everywhere in the UI

fernandojsg:
Quite old and I'm still missing while playing with it, merging! and if we want to modify shortcuts we could do later on.

Issue #317, opened 2016-08-22
Support adding of components


fernandojsg:
Following up with the `Add component (+)` icon proposed by @feiss in https://github.com/aframevr/aframe-inspector/issues/316 I would like to hear some opinions about how to deal with it.

Right now we've just a combobox that is populated with the core's components and the third party components loaded from https://github.com/aframevr/aframe-components.

For the third party components we've a mockup of how to deal with them, in a similar way as unity's store where users can browse them:
![mo_v2_components-modal](https://cloud.githubusercontent.com/assets/782511/17850326/fe436222-685c-11e6-8bab-23cbf01f7fa0.png)
I like this kind of dialog to promote components discovery by the users, they could check around, search for a custom component, visit its website to see more examples, see a screenshot, etc...
But I don't believe it's handy to keep opening this dialog every time that the user clicks `Add component`.

I like the way Unity handles this. When you click `Add component` there you get a dialog with a searchbox focused by default and the list of available components:
<img width="231" alt="screenshot 2016-08-22 11 41 00" src="https://cloud.githubusercontent.com/assets/782511/17850406/54e7639e-685d-11e6-9e97-b1acee512b18.png">

We could have a similar behaviour as Unity:
1. Showing the list of available components (core and loaded thirdparty components) with a searchbox and a button to launch the "components store" to include new components on the scene (they'll be available also on the list of available components).
2. The same as before but all the third party components will be available in the components list. You could enter the components dialogs just to take a look at the extra info.

feiss:
I like the first option, showing a dropdown list with a searchbox, and a link to the store

feiss:
'Store' is a term I dislike if all components will be for free (if not, it's fine), btw.

fernandojsg:
Yep, I used `store` just to get the idea of `Asset store` from unity but won't be that the name in the UI.
Ok, so I'll try to mockup it on the inspector.

fernandojsg:
A WIP:
![addcomponent](https://cloud.githubusercontent.com/assets/782511/17860914/6b424410-688f-11e6-9c73-b27bc81d59af.gif)
If we're going to switch to the "3dsmax style" to show the components we'll just select the new component on the list and show its properties so the users will have some feedback.
With the current UI we should probably add some animation or text `Component "Sound" Added` as the user may cannot see the new component added to the bottom of the list.

fernandojsg:
[...]

PR #325, opened 2016-09-25
Added basic component-registry loader

fernandojsg:
Based on the latest changes from https://github.com/aframevr/aframe-registry
screen shot 2016-09-25 at 16 14 16


fernandojsg:
I'm not sure about which name to use for grouping the components. At first I thought about "core" and "registry/community" but if they user already has some custom components loaded they'll appear on "core" too as it shows really the names of the components already loaded.
Any thoughts on that @ngokevin ?

ngokevin:
For now, it could just be `registered` + `community` or `registry`. 

Though you can also check components against the `registry` to determine whether or not they are core, developer-registered, or community.

But not too important since the dropdown is just temporary right?

fernandojsg:
I've just changed it to `registered` + `registry`. The dropdown as it's right now it will be just temporary but still we'll have a autocomplete dropdown where you can access these components, as discussed on https://github.com/aframevr/aframe-inspector/issues/317: https://cloud.githubusercontent.com/assets/782511/17860914/6b424410-688f-11e6-9c73-b27bc81d59af.gif

Next to this dropdown you'll have a components button that will popup the big dialog similar to Unity's store. You'll use that dialog when you want to know more info about a component, or when you want to discover new components as it will show some descriptions, links to author/components, tags, and so on.
But still we want to keep the autocomplete-dropdown as it's faster to include already known and commonly used components such as `geometry`, `material` and so on.

src/components/components/AddComponent.js
```diff
@@ -13,39 +14,68 @@ export default class AddComponent extends React.Component {
...
+      Events.emit('componentAdded', {entity: entity, component: componentName});
+      ga('send', 'event', 'Components', 'addComponent', componentName);
     }
 
-    entity.setAttribute(componentName, '');
-    Events.emit('componentAdded', {entity: entity, component: componentName});
-    ga('send', 'event', 'Components', 'addComponent', componentName);
+    if (origin === 'community') {
+      var packageName = selectedOption.value;
+      INSPECTOR.componentLoader.addComponentToScene(packageName, function (componentName) {
```
ngokevin: can one line `INSPECTOR.componentLoader.addComponentToScene(packageName, addComponent)`

src/lib/componentloader.js
```diff
@@ -7,29 +7,29 @@ ComponentLoader.prototype = {
   loadComponentsData: function () {
     var xhr = new window.XMLHttpRequest();
     // @todo Remove the sync call and use a callback
-    xhr.open('GET', 'https://raw.githubusercontent.com/aframevr/aframe-components/master/components.json', false);
+    xhr.open('GET', 'https://aframe.io/aframe-registry/build/' + AFRAME.version + '.json');
```
ngokevin: you have to restrict this to the major version:
- if it's 0.3.x, use 0.3.0
- if it's 4.x.x, use 4.0.0

ngokevin (approved):

src/components/components/AddComponent.js
fernandojsg: Yep! I missed it :D

src/lib/componentloader.js
fernandojsg: I've just added the function `getMajorVersion (version)` that takes cares of that

PR #329, opened 2016-10-01
Fixes #328 show recursively elements in scenegraph even if they're no…

### LINKED ISSUE #328: primitives missing from scene graph
Open inspector on https://samsunginternet.github.io/a-frame-demos/racer/

Under each track, there should be `<a-curve-point>` primitives.

``` html
 <a-curve id="track1" curve="CatmullRom">
        <a-curve-point position="0 -4 0">
            <a-curve-point position="-30 8 10">
                <a-curve-point position="-30 -6 30">
                    <a-curve-point position="0 8 40">
                        <a-curve-point position="0 -2 40">
                            <a-curve-point position="0 25 50">

                            </a-curve-point>
                        </a-curve-point>
                    </a-curve-point>
                </a-curve-point>
            </a-curve-point>
        </a-curve-point>
    </a-curve>
```

fernandojsg:
…t a-entity

### DIFF
--- a/src/components/scenegraph/SceneGraph.js
+++ b/src/components/scenegraph/SceneGraph.js
@@ -120,9 +120,6 @@ export default class SceneGraph extends React.Component {
             html: html
           });
 
-          if (child.tagName.toLowerCase() !== 'a-entity') {
-            continue;
-          }
           treeIterate(child, depth);
         }
       }


ngokevin:
You can check child.isEntity

ngokevin:
r+wc

fernandojsg:
Yes, I was checking that already https://github.com/aframevr/aframe-inspector/blob/master/src/components/scenegraph/SceneGraph.js#L97 
I just didn't iterate recursively if the entity was a-animation or so.

PR #330, opened 2016-10-02
Fix circular dependencies for INSPECTOR

fernandojsg:
Before we were exporting `DEFAULT_COMPONENTS` in `CommonComponents.js` this generates and odd error due to circular dependencies when importing `INSPECTOR` in `TextureWidget.js`.
I've fixed it by moving `DEFAULT_COMPONENTS` to a new file `DefaultComponents.js`.


ngokevin:
yeah with circ deps, you have to decouple or else it'll return null on require

PR #336, opened 2016-10-07
support component groups from registry

ngokevin:
There are some kinks, but at least the component/systems register + work.

https://github.com/aframevr/aframe-registry/pull/26


fernandojsg:
I've tried it and when I want to add the component `static-body` to an entity it will add also `dynamic-body` is it an expected behaviour for that component? Because it seems strange

ngokevin:
Fixed. It was late so I messed up. Now passes the component name it wants to inject, and only registers that. Also made it a Promise.

fernandojsg:
Cool!

Issue #316, opened 2016-08-19
Faulty visual hierarchy in components panel


feiss:
_Disclaimer: I'm new in town, so please bear with me about all these suggestions I'm throwing. I'm not familiar with all the discussions, prototypes, past versions and motivations of this topics regarding the editor, I'm just giving my naive opinion on what I see in current version._

Folowing @cvans 's issue on #311 , I feel the hierarchy of the headers in the components panel could be better.

Now:

![image](https://cloud.githubusercontent.com/assets/359872/17814110/7f15e662-662e-11e6-96f7-d6d140bba895.png)

Issues:
- `<a-entity>` is almost unseen, and really is the main character here! The whole panel is about this guy, and should be H1
- Since all entities will have the "common" section, I find it useless. Also, "common" is not a name of an actual component, so it's a bit misleading. I'd just remove it, leaving the common properties just below of the entity header.
- The same header and style is used for adding a new component ("Components") and the actual component, when the semantics of both are completely different. I propose removing the "Components" header and replace the select for a button (that opens a floating list of components, or the components gallery)
- Two different 'copy html' buttons
- Some properties have help button and others don't. This could be another issue itself, but I suggest removing all these buttons and use a more subtle help system, like:
  - The actual property name is a link to its doc (simple implementation), or
  - A main help button that changes the mouse pointer and allows the user clicking any header/title/property/field of the editor and open the help/spec/doc link associated to it (not quite simple).
- There's no clear meaning in the use of colors: some user entry fields are in blue, and other (text) in gray. Also, some buttons are grey and others are blue. It would be nice to have a clear relationship between colors and function, for example, blue for user entry fields, and gray (or pink/magenta, like the "Back to Scene" button) for the buttons.

I did other minor tweaks too, but not for discussion in this issue I guess. The main topic I wanted to arise is the faulty visual hierarchy in the elements of the panel and the use of the section headers.

![image](https://cloud.githubusercontent.com/assets/359872/17814876/10c12a1a-6632-11e6-8c8c-a85d5f57b738.png)

fernandojsg:
We love feedback so don't need to be worried about it, it's so cool and useful to have an open conversation about the app :+1:

> <a-entity> is almost unseen, and really is the main character here! The whole panel is about this guy, and should be H1

Yep totally right something bigger and brighter should be used for that.

[...]

PR #339, opened 2016-10-14
New sidebar

fernandojsg:
- Added `react-select` for selectboxes, and style them
- Style sidebar according to the ideas from @feiss (https://github.com/aframevr/aframe-inspector/issues/316 - _Still a lot to do but will address that in another PR._):
  - Bigger sidebar title
  - `Add component` is _part of_ the common attributes panel, but if you collapse commons, it `Add component` will still be visible.
  - `Add component` replaced previous selectbox with an autocomplete selectbox.
- Fixes a problem when scrolling sidebar that it prevented us to see the latests attributes.

newsidebar

### DIFF
--- a/src/components/components/AddComponent.js
+++ b/src/components/components/AddComponent.js
@@ -15,14 +15,13 @@ export default class AddComponent extends React.Component {
    * If component is instanced, generate an ID.
    */
   addComponent = (componentName) => {
-
     var entity = this.props.entity;
     var select = this.refs.select;
     var selectedOption = this.options.filter(function (option) {
       return option.value === componentName;
     })[0];
 
-    if (origin === 'registry') {
+    if (selectedOption.origin === 'registry') {
       var [packageName, componentName] = selectedOption.value.split('.');
       INSPECTOR.componentLoader.addComponentToScene(packageName, componentName)
         .then(addComponent);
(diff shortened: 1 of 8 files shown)


src/components/components/AddComponent.js
```diff
@@ -65,45 +68,38 @@ export default class AddComponent extends React.Component {
...
-                </optgroup>
-              </select>
-              <a className='button fa fa-plus-circle' onClick={this.addComponent}/>
-            </span>
-          </div>
+        <div className='add-component2'>
+          <Select
+            className="add-component"
+            ref="select"
+            autofocus
```
ngokevin, line 93: should it autofocus? right now it focuses if you go from no entity selected to entity selected. but switching between entities, it doesn't autofocus

ngokevin:
can we distinguish between registry components?

ngokevin:
r+wc

src/components/components/AddComponent.js
fernandojsg: I did a fix by using

``` js
  componentDidUpdate (nextProps) {
    this.refs.select.focus();
  }
```

But I removed it because it prevents the scenegraph from taking the focus, so if you click on the scenegraph and try to use the arrow keys to move up and down it doesn't affect the scenegraph but the selectbox.
I've created another issue (https://github.com/aframevr/aframe-inspector/issues/342) to discuss some ideas on how to improve it, maybe detect that you've clicked the objects in the viewport and focus the add component, but if you click on the scenegraph just keep the focus there.

PR #344, opened 2016-10-14
Added help modal for shortcuts

fernandojsg:
It shows up when pressing `h` and close by clicking `X` or pressing `esc`

helpmodal

It also fixed a bug as we didn't call `stopPropagation` when listening for `esc` key on modals, and we were closing the modal and exiting the inspector at the same time.

Restyled based on @feiss proposal:
image

### DIFF
--- a/src/components/Main.js
+++ b/src/components/Main.js
@@ -6,6 +6,7 @@ import ReactDOM from 'react-dom';
 
 const Events = require('../lib/Events.js');
 import ComponentsSidebar from './components/Sidebar';
+import ModalHelp from './modals/ModalHelp';
 import SceneGraph from './scenegraph/SceneGraph';
 import ToolBar from './ToolBar';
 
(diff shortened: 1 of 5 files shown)


feiss:
why not closing it with `h` too instead (or apart from) of `esc`? I'd expect that `h` works as a toggle

fernandojsg:
@feiss Done!, I've added a `extraCloseKeyCode` as optional parameter to modal, and defined `h` for the help modal. It will still close if you press `esc` as most people expect that to works, but `h` too.

dmarcos:
Should there be also a button anywhere to toggle help?

fernandojsg:
@dmarcos Not sure, by now I wouldn't care to add this one for the shortcuts, as I would expect from a help button somewhere on the UI to act as help for the whole inspector. Maybe we could do later a PR modifying this shortcuts modal and add in the top some little text with the links to the inspector and aframe repo or so, and then we could add that help button to the UI

dmarcos:
it looks good to me. We can rethink help a bit later.

fernandojsg:
Cool, thanks!

PR #346, opened 2016-10-14
Fixes #321 Changing editor camera near parameter

### LINKED ISSUE #321: Near limit of camera's frustum cuts small objects
When inspecting small objects (like less than 1,1,1), camera's frustum limits cuts the object.

Cube 0.5, 0.5, 0.5

image

I propose decrease the limit a bit more. It's interesting to get pretty close to objects when you are reviewing texture or material issues.

fernandojsg:
... so you can get much closer to the objects
image

### DIFF
--- a/src/lib/inspector.js
+++ b/src/lib/inspector.js
@@ -50,7 +50,7 @@ Inspector.prototype = {
       this.EDITOR_CAMERA = this.inspectorCameraEl.getObject3D('camera');
       this.initUI();
     });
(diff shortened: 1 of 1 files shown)


dmarcos:
I would even make it 0.001. It will be important for VR mode. 0.1 means that objects will be clipped 10 cm from your eyes.

fernandojsg:
Yes, the VR mode could have another setup for the camera. I would go for something in the middle 0.01, going to something much smaller like 0.001 will result in losing resolution in the depth buffer and we could get some z-fighting.

PR #352, opened 2016-10-17
Fixes #349: Newly created entities were no selectable from the viewport

### LINKED ISSUE #349: Newly created entities with geometry are not selectable in the viewport

fernandojsg:
- Allow newly created entities to be selected on the viewport aswell as its helpers
- Start using `componentinitialized` for newly created components.
- Created `empty.html` to use for testing empty scenes

### DIFF
--- a/example/empty.html
+++ b/example/empty.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Example Scene</title>
+    <script src="js/aframe.min.js"></script>
+  </head>
+  <body>
+    <a-scene stats>
+    </a-scene>
+    <script src="http://localhost:3333/build/aframe-inspector.js"></script>
+  </body>
+</html>
--- a/src/components/components/Sidebar.js
+++ b/src/components/components/Sidebar.js
@@ -35,13 +35,20 @@ export default class Sidebar extends React.Component {
     Events.emit('selectedentitycomponentchanged', event.detail);
   }
 
+  componentCreated = (event) => {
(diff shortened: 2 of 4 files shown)


example/empty.html
```diff
@@ -0,0 +1,13 @@
...
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Example Scene</title>
+    <script src="js/aframe.min.js"></script>
+  </head>
+  <body>
+    <a-scene stats>
+    </a-scene>
+    <script src="http://localhost:3333/build/aframe-inspector.js"></script>
```
dmarcos, line 11: Is this to have an empty scene to edit?

src/components/components/Sidebar.js
```diff
@@ -35,13 +35,20 @@ export default class Sidebar extends React.Component {
     Events.emit('selectedEntityComponentChanged', event.detail);
   }
 
+  componentCreated = (event) => {
+    Events.emit('selectedEntityComponentCreated', event.detail);
```
dmarcos, line 39: Should we remove camel case from event names for consistency?

example/empty.html
fernandojsg: Yes, many times I want to try the inspector with an empty scene and it's annoying to edit the current example or delete every element.

src/components/components/Sidebar.js
fernandojsg: Yep, I've opened an issue for that https://github.com/aframevr/aframe-inspector/issues/357

fernandojsg: And here you have it: https://github.com/aframevr/aframe-inspector/pull/358

ngokevin:
r+

PR #356, opened 2016-10-18
Fix #354 implement vec2 widget

### LINKED ISSUE #354: Add vec2 widget

fernandojsg:
While implementing it I found a bug that prevents the compound widgets like `vec3` to work properly outside of a single property component (Like `position`, `rotation`, ...).

I simplified the `updateEntity` function (Probably it will be a good idea to open an issue/PR on aframe to use it or modify `AFRAME.utils.entity.setComponentProperty` to accept `null/undefined` value.

I simplified the `onChange` callback on each widget as is not necessary anything else than the `value` and actually the `attribute name` (x,y,z,...) for compounds widgets.

I was thinking about leaving the tests for another PR as I'm not that familiar with this test system and as this is broken right now so we could get it working and then focus on testing side?

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -6,7 +6,7 @@
     <script src="js/aframe.min.js"></script>
   </head>
   <body>
-    <a-scene>
+    <a-scene debug="true">
       <a-assets>
         <a-mixin id="blue" material="color: #4CC3D9"></a-mixin>
         <a-mixin id="blueBox" geometry="primitive: box; depth: 2; height: 5; width: 1" material="color: #4CC3D9"></a-mixin>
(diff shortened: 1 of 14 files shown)


example/index.html
```diff
@@ -6,7 +6,7 @@
     <script src="js/aframe.min.js"></script>
   </head>
   <body>
-    <a-scene stats>
+    <a-scene stats debug="true">
```
ngokevin: need?

ngokevin:
r+

example/index.html
fernandojsg: I like to have it enable by default as I use the scene to test the inspector itself and It's nice to see how values are changing. But maybe we could just inject it by default once the inspector is opened? and remove when closing?

fernandojsg: I've opened and issue so we could discuss it there https://github.com/aframevr/aframe-inspector/issues/367

PR #365, opened 2016-10-19
Improved svg aframe-registry logo

feiss:
- Improved reability of logo in small size
- Removed external svg file and embedded the svg in the html. Now it's much more optimized and its properties (color) can be modified via css


src/components/components/AddComponent.js
```diff
@@ -100,7 +100,9 @@ export default class AddComponent extends React.Component {
             searchable={true}
           />
           <a href="https://aframe.io/aframe-registry" target="_blank" title="A-Frame Registry" className="aregistry-button">
-            <img src="../assets/a-registry-logo-min.svg"/>
+            <svg version="1.1" viewBox="0 0 19.999999 19.999999">
```
fernandojsg, line 103: It's quite ugly to have it hardcoded here isn't it? Why can't we just use an external svg so if we want to reuse somewhere else, like in the registry modal or whatever we could reuse it, or change it easly by updating just the .svg?

feiss:
closed and replaced by #368

Issue #297, opened 2016-08-18
Style scrollbar


fernandojsg:

cvan:
Are you suggesting hijacking the native scrollbar/scrolling? If you're suggesting just pushing it to the side so it's not visible, and maybe showing a custom one, that's fine.

fernandojsg:
I think so, I don't have much experience styling scrollbars so any help will be more than welcome. Current scrollbar style ruins the UI quite a bit :)

cvan:
Check out http://theculturetrip.com/europe/greece/articles/spotlight-on-the-stavros-niarchos-cultural-center/ or http://www.apple.com/mac/ at a mobile width for an example. There's a little hackery involved (you need a parent element, need to add some extra padding, and then clever cosmetic styling to simulate the scrollbar).

cvan:
Also, I'm sure there is at least _one_ proper solution on CodePen: https://codepen.io/search/pens?q=scrollbar&limit=all&type=type-pens

ngokevin:
wontfixing for now. only chrome supports, and don't feel like maintaining hacks / faked scrollbars

PR #371, opened 2016-10-19
added styling for scroll-bar

odGit:
This it an initial changes for issue #297
I would like to hear you feedback and maybe some sketch of how you would like it.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -923,3 +923,26 @@ span.entity-name {
   width: 20px;
   height: 20px;
 }
+
+/*styling of a scroll-bar*/
+::-webkit-scrollbar {
+  width: 14px;
+}
+
+::-webkit-scrollbar-thumb {
+  border: 4px solid rgba(0, 0, 0, 0);
+  background-clip: padding-box;
+  -webkit-border-radius: 7px;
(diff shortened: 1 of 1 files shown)


fernandojsg:
Hi @odGit ! thanks for helping with that.
I don't have much experience with styling scrollbars, but is not possible to style it also on firefox? I've see it working just on chrome.
About the style, I believe @feiss could help here giving some tips as he's taking care of the overall inspector aesthetics.

feiss:
afaik, the only way of styling scrollbars in firefox is not using native scrollbars, but a js library like https://www.npmjs.com/package/react-scrollbar

odGit:
Thanks @feiss, I will have a look at this one.

fernandojsg:
It looks like they didn't update it for a while and it has many bugs, some of them related to dynamically modify the height of the container. So I don't feel quite confident about using it.
Maybe we could just define an style for chrome and leave firefox for another PR if we find a more stable scrollbar solution.

ngokevin:
https://github.com/aframevr/aframe-inspector/commit/c49af2ed2adcc674d1e2e4354c149ea462d9102f

Issue #304, opened 2016-08-18
User can't be used under the toolbar


fernandojsg:
Because of the use of `position: fixed` for the right sidebar, it overlaps the canvas so it doesn't let it get the mouse events 👎 
<img width="1023" alt="screenshot 2016-08-18 19 02 43" src="https://cloud.githubusercontent.com/assets/782511/17783010/5e3c4be2-6576-11e6-8f60-e194f0d27dd9.png">

fernandojsg:
Closed by https://github.com/aframevr/aframe-inspector/pull/372

PR #372, opened 2016-10-19
Fix overlap right sidebar in to canvas

nucliweb:
Fix issue #304 
- Remove `bottom: 0;` at css rule _#right-panels_


fernandojsg:
Wow!!! Thank you so much for fixing this ugly bug :)

PR #379, opened 2016-10-25
added new icons for copy html and delete component

feiss:


fernandojsg:
What about converting the svg to a font instead? With an app like https://icomoon.io/app I could give it a try. As we're going to have many icons it could be more useful

dmarcos:
I've used icomoon in the past. It's good.

fernandojsg:
@feiss what about adding the icons using `iconomoon` and add the font to the source to the repo so we could keep adding the images there incrementally so we could remove fontawesome icons in the future

feiss:
yup, I'd switch to iconmoon

ngokevin:
outdated, feiss will probably do in a bit in a new pr

PR #380, opened 2016-10-25
simplify inspector usage instructions

ngokevin:
Promote the keyboard shortcut as the primary way to use the Inspector by reducing other clutter instructions. Some people glossed over the shortcut and were reading about including a build.


fernandojsg:
:+1:

PR #392, opened 2016-11-18
Fixes #391 - Inspector panel doesnt works if registry build fails

### LINKED ISSUE #391: If the registry build fails to fetch, Inspector component panel doesn't work
INSPECTOR.componentLoader.components is null, and Object.keys fails.

fernandojsg:


src/lib/componentloader.js
```diff
@@ -22,8 +22,12 @@ ComponentLoader.prototype = {
     xhr.open('GET', registryBaseUrl + getMajorVersion(AFRAME.version) + '.json');
 
     xhr.onload = () => {
-      this.components = JSON.parse(xhr.responseText).components;
-      console.info('Components in registry:', Object.keys(this.components).length);
+      if (xhr.status === 404) {
+        console.error('Error loading registry file: 404');
+      } else {
+        this.components = JSON.parse(xhr.responseText).components;
+        console.info('Components in registry:', Object.keys(this.components).length);
```
dmarcos: Do we want a better message here? The error message shows what happened but success does not.
"Registry file loaded. " + Object.keys(this.components).length + " components available"

dmarcos:
Just made a small comment on the success message. It looks good to me.

src/lib/componentloader.js
fernandojsg: Thanks, modified!

PR #393, opened 2016-11-18
Fixes #312 Copy HTML

### LINKED ISSUE #312: Issues with "Copy HTML" functionality (partially broken?)
image

**Problem:**
It appears that only when values _change_ can I `copy html`. So, upon immediately selecting an entity, I clicked on the `Geometry` component's `copy html` button, and it added nothing to my clipboard. When I checked the `skipCache` checkbox, my clipboard now contained `skipCache:true`. But, when I changed the `primitive`, now my clipboard was empty again. I'm confused.

**Proposed solutions:**
- Hide "Copy HTML" button if there are no changes to make.
- Rename "Copy HTML" to "Copy HTML changes" (or "patches" as Vizor calls them).
- Change the functionality so that "Copy HTML" for a Component always copies the entire HTML. (I assume it's on the roadmap eventually to enable granular copy-and-paste functionality for each property? I saw some discussion in issue #301 filed by @feiss. Should a separate issue be filed? This seems important to either allow "Copy all HTML to clipboard" or enable granular singular/multiple copying à la the checkboxes that appear in browsers' Dev Tools' Style manipulation tools).

<hr>

**Problem:**
I change a few values on the `Material` component. I click `copy html` and the resulting text in my clipboard is this:

``` css
flatShading:true;transparent:true;color:#F16745
```

**Proposed solution:**
Can we add spaces between the `:` and `;` delimiters so it's more readable?

I was thinking perhaps actually the output should contain the originating parent Component, like so:

``` html
material="flatShading: true; transparent: true; color: #F16745"
```

But, then I realised that a developer may accidentally replace the other properties in the Component that weren't included

<hr>

**Problem:**
Without modifying any values, if I select an entity and immediately press its `click html` button, the Inspector thinks I modified the colour and this appears in my clipboard:

``` css
color:#F16745
```

**Proposed solution:**
Fix whatever's causing `color` to think it's been changed.

fernandojsg:
As proposed in https://github.com/aframevr/aframe-inspector/issues/312:
* Replaced `Copy HTML` with `Copy attributes`
* Added spaces after `;` and `:` to make it easier to read:
  * before: `primitive:cylinder;height:0.2;radius:12`
  * now: `primitive: cylinder; height: 0.2; radius: 12`
* Added component's name and double quotes to the output: 
  * before: `primitive:cylinder;height:0.2;radius:12`
  * now: `geometry="primitive: cylinder; height: 0.2; radius: 12"`

### DIFF
--- a/src/actions/component.js
+++ b/src/actions/component.js
@@ -32,7 +32,8 @@ function getModifiedProperties (entity, componentName) {
  */
 export function getClipboardRepresentation (entity, componentName) {
   var diff = getModifiedProperties(entity, componentName);
-  return AFRAME.utils.styleParser.stringify(diff);
+  var attributes = AFRAME.utils.styleParser.stringify(diff).replace(/;|:/g, '$& ');
+  return `${componentName}="${attributes}"`;
 }
 
 /**
--- a/src/components/components/Component.js
+++ b/src/components/components/Component.js
@@ -109,7 +109,7 @@ export default class Component extends React.Component {
             <a title='Copy to clipboard' data-action='copy-component-to-clipboard'
               data-component={subComponentName || componentName}
               className='flat-button' onClick={event => event.stopPropagation()}>
-              Copy HTML
+              Copy Attributes
             </a>
             <a title='Remove component' className='flat-button'
               onClick={this.removeComponent}>Remove</a>


dmarcos:
Tried the regular expression and it works as advertised:
image

It looks good to me.

fernandojsg:
Thanks!

PR #396, opened 2016-11-18
Fixes #340 Differentiate between registry and core components in Add …

### LINKED ISSUE #340: Differentiate between registry and core components in Add Component selectbox
We'll have three types of components:
- Core
- Registry
- Manually added to the html

We should find a way to differentiate them on the `Add component` select box, maybe different icon or color depending its origin?

fernandojsg:
…Component selectbox

+ sort the list of components alphabetically
+ added tooltip to the already-loaded bullet

image


dmarcos:
It looks good to me.

PR #397, opened 2016-11-18
Fixes build/dist

fernandojsg:
We had `/build` and `/dist` folders that were being used differently using webpack devserver and when deploying and it was a quite messy. 
So I've just removed any reference to `/build' and everything will pointing to `/dist`


ngokevin:
r+, also can remove the old build folder

Issue #2, opened 2016-06-30
Remove invalid component's attributes


fernandojsg:
If we have an entity with an standard material like:

``` html
<a-entity geometry="primitive: sphere" material="metalness:0.0;roughness:1.0;shader:standard"></a-entity>
```

And we change the shader, some of the parameters won't be available on this new shader (metalness, roughness):

``` html
<a-entity geometry="primitive: sphere" material="metalness:0.0;roughness:1.0;shader:noise"></a-entity>
```

So probably we should delete them, as it will keep giving warnings on the console:

```
core:schema:warn Unknown component property: metalness +0ms
browser.js:112 core:schema:warn Unknown component property: roughness +2ms
```

fernandojsg:
@dmarcos could you take a look at it? As I believe it may be related with the work you did on the copy to clipbard

dmarcos:
This warning is expected since the schema has changed. I'm not super fan of removing any attributes since it might surprise the user.

fernandojsg:
Yes I'm not sure about removing the attributes either, I feel like it's nice as you're just removing attributes not used with the current schema, but on the other hand it's true that you could delete an attribute that the user could need later, like having a `sphere` with `radius=10`, switching to `box` by mistake and then back to `sphere` and the `radius` value is gone...
@ngokevin what do you think? better to close it?

ngokevin:
Yeah, I'd close. Wouldn't mess with user code. Maybe they expect to change schema later or something.

fernandojsg:
Cool, thanks!

PR #398, opened 2016-11-19
Remove attributes that represent mixins and have not changed with res…

### LINKED ISSUE #310: Should "Copy HTML" for an entity include unaltered/empty components?
For example, see `material="" geometry=""`:

``` html
<a-entity id="shortOrangeBox" mixin="short orange box" position="-5 2 0" material="" geometry=""></a-entity>
```

dmarcos:
…pect

to the mixed in value (fix #310)

### DIFF
--- a/src/actions/entity.js
+++ b/src/actions/entity.js
@@ -2,6 +2,7 @@ var Events = require('../lib/Events.js');
 var components = AFRAME.components;
 var isSingleProperty = AFRAME.schema.isSingleProperty;
 
+import {equal} from '../lib/utils.js';
 import DEFAULT_COMPONENTS from '../components/components/DefaultComponents';
 
 /**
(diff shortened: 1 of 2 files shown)


fernandojsg:
I've just tried to copy html the default cylinder from the example and I got the following result, please notice that `position` is generated even if it's the default values so it shouldn't be there:

```
<a-entity id="floor" geometry="primitive:cylinder;height:.2;radius:12" material="src:#floorImg;color:#fafafa;metalness:.2;repeat:50 20;roughness:.1" position="0 0 0">
</a-entity>
```

If I change the `primitive` to `box` that attribute shouldn't appear on the generated HTML as it's using the default value for that attribute:

```
<a-entity id="floor" geometry="primitive:box;height:.2;radius:12" material="src:#floorImg;color:#fafafa;metalness:.2;repeat:50 20;roughness:.1" position="0 0 0"></a-entity>
```

In the same way, if I change the geometry component to set the default values for `primitive`, `height`, `depth` and `width`, I would expect to copy just `geometry=""`:

```
<a-entity id="floor" geometry="primitive:box;height:1;radius:12;depth:1;width:1" material="src:#floorImg;color:#fafafa;metalness:.2;repeat:50 20;roughness:.1" position="0 0 0"></a-entity>
```

Although `radius` attribute should be removed in another PR to fix the issue: https://github.com/aframevr/aframe-inspector/issues/2

PR #405, opened 2016-12-22
Properly sort the components in AddComponents widget

jsantell:
`Array.prototype.sort` does not destructively modify the array and expects a number, not a boolean


fernandojsg:
@jsantell Cool! I didn't know that, thank you!

PR #410, opened 2016-12-23
Do not fire hotkey commands based off of active element, but prevent

### LINKED ISSUE #406: Gimbal mode not always changable via hotkey
When I hit the hot keys (W, E, R) to change the position, rotation or scale in the viewport, even though the gimbal is visible, the mode does not change. For example, select an entity to display the gimbal -- when it first appears upon a new selection, the hotkeys do not work. Selecting the components view for focus enables the functionality. When selecting an entity, and there's no previous entity selected, focus appears in the "Add Component..." filter, causing hotkeys to be entered as a component filter.

jsantell:
input elements from propagating events to document if they should not be
acted on. Fixes #406

### DIFF
--- a/src/lib/shortcuts.js
+++ b/src/lib/shortcuts.js
@@ -1,7 +1,12 @@
 /* globals AFRAME */
 var Events = require('./Events');
-var shouldCaptureKeyEvent = AFRAME.utils.shouldCaptureKeyEvent;
 import {removeSelectedEntity, cloneSelectedEntity} from '../actions/entity';
+
+function shouldCaptureKeyEvent (event) {
+  if (event.metaKey) { return false; }
+  return event.target.tagName !== 'INPUT' &&
+    event.target.tagName !== 'TEXTAREA';
+}
 
 module.exports = {
   onKeyUp: function (event) {


jsantell:
Using the method described in #406 -- any issues you think with this approach? Or any input field areas that aren't accounted for? Works best alongside #407 so that hotkeys are also usable when selecting the first entity

fernandojsg:
What about, instead of including `stopPropagating` on every possible input control, modify the code for `shouldCaptureKeyEvent` for the inspector?
Something like:
```
function shouldCaptureKeyEvent(event) {
  if (event.metaKey) { return false; }
  return document.activeElement === document.body &&
    (event.target.tagName !== 'INPUT' ||
    event.target.tagName !== 'TEXTAREA');
}
```

jsantell:
@fernandojsg That would work as well! My only concern would be interactions like arrow keys moving the camera around when the Entities panel has focus (would scroll the panel instead), but seems like a separate issue of handling focus (and not sure which behaviour is even preferred!) -- all in all, I think the solution of checking for origin of the event and canceling on input elements would be the smallest change for now!

jsantell:
That being said, I believe `document.activeElement === document.body` is the part that is currently preventing shortcuts from occuring when the entities panel has focus, as the SceneGraph's .outliner div steals focus

jsantell:
Pushed a change based on feedback

fernandojsg:
Cool! thank you

PR #413, opened 2016-12-29
Fixes #404 Color input

### LINKED ISSUE #404: CSS color names appear as black in ColorWidget
When setting a color property to a CSS color like 'white' or 'orange', the color widget displays it as black. The widget handles converting from the short hand hex variant `#333`, but does not handle color names, or HSL values.

I can write a quick patch for this, but wondering what's more desirable -- using a lib like [color-string](https://www.npmjs.com/package/color-string) (+14kb unminified, +500b minified), or putting the supplied color value into a div and using `window.getComputedStyle` and converting by hand the supplied HSL values. Both methods should handle any color format.

fernandojsg:
- Added input box
- Added support for CSS color format: rgb, hsl, hex3, hex6, color names.

/cc @dmarcos @jsantell

### DIFF
--- a/src/components/widgets/ColorWidget.js
+++ b/src/components/widgets/ColorWidget.js
@@ -15,34 +15,69 @@ export default class ColorWidget extends React.Component {
 
   constructor (props) {
     super(props);
-    this.state = {value: correctColorValue(this.props.value) || ''};
+
+    var value = this.props.value;
+    this.color = new THREE.Color();
+
+    this.state = {
+      value: value,
+      pickerValue: this.getHexString(value)
+    };
   }
 
-  componentWillReceiveProps (newProps) {
-    if (newProps.value !== this.state.value) {
-      this.setState({value: correctColorValue(newProps.value)});
-    }
-  }
+  setValue (value) {
+    var pickerValue = this.getHexString(value);
 
-  onChange = e => {
-    var value = e.target.value;
-    this.setState({value: value});
+    this.state = {
+      value: value,
+      pickerValue: pickerValue
+    };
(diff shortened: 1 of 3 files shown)


src/components/widgets/ColorWidget.js
```diff
@@ -15,34 +15,77 @@ export default class ColorWidget extends React.Component {
...
+
   onChange = e => {
-    var value = e.target.value;
-    this.setState({value: value});
-    if (this.props.onChange) {
-      this.props.onChange(this.props.name, value);
+    this.setValue(e.target.value);
+  }
+
+  onBlur = e => {
```
jsantell: `onBlur` is not used

src/components/widgets/ColorWidget.js
```diff
@@ -15,34 +15,77 @@ export default class ColorWidget extends React.Component {
...
+          value={this.state.value}
+          onKeyUp={this.onKeyUp}
+          onChange={this.onChangeText}
+        />
+      </span>
     );
   }
 }
 
 function correctColorValue (value) {
```
jsantell: `correctColorValue` is no longer used

jsantell (commented):
Just pulled down the changes and they work great!

src/components/widgets/ColorWidget.js
fernandojsg: Fixed

src/components/widgets/ColorWidget.js
fernandojsg: Fixed! thanks

Issue #409, opened 2016-12-22
"Copy Entity HTML to Clipboard" does not work in the example


jsantell:
Haven't dug in why it's only some entities that throw errors, but in the built in example (and some entities in my own scenes and the aframe examples), a few entities throw when copying:

`<a-scene />`
```
Uncaught TypeError: Cannot read property 'length' of null
    at removeDefaultAttributes (entity.js:156)
```

`<a-box />`
```
Uncaught TypeError: Cannot read property 'getAttribute' of null
    at Object.selectorStringify [as stringify] 
```

Have also seen errors with `addEventListener` and `removeChild` when attempting to copy, but on other entities it works fine.

fernandojsg:
@dmarcos could you take a look at it? It looks like a problem on `getClipboardRepresentation` when calling `flushToDOM` here: 
https://github.com/aframevr/aframe-inspector/blob/master/src/actions/entity.js#L144

fernandojsg:
I've just updated to the latest aframe in the example and it seems to fix some of the issues, or at least modify the error :)
So far it works on entities that doesn't have maps, but in the ones that have maps like `a-box` or `floor` it throws:
```
aframe.min.js:275 Uncaught TypeError: Cannot read property 'systems' of undefined
    at Object.module.exports.updateMap (aframe.min.js:275)
    at t.init (aframe.min.js:240)
    at i.updateShader (aframe.min.js:79)
    at i.update (aframe.min.js:79)
    at i.updateProperties (aframe.min.js:145)
    at HTMLElement.value (aframe.min.js:137)
    at HTMLElement.value (aframe.min.js:137)
    at HTMLElement.value (aframe.min.js:137)
    at HTMLElement.value (aframe.min.js:143)
    at HTMLElement.value (aframe.min.js:143)
```

dmarcos:
This one is trickier than I thought and I'm a bit surprised we did not see the problem before. The issue is that `updateMap` relies on `sceneEl` being available. The `Copy Entity HTML to Clipboard` feature clones the selected entity to avoid undesired modifications and manipulates it to remove default and mixed in values. The cloned entity is not part of `a-scene` so `updateMap` does not find the reference it needs. The problem is that `updateMap` is invoked from the `update` method which should not ever be called for a not yet loaded entity. I have to investigate a bit more to figure out a proper solution.

dmarcos:
[...]

PR #418, opened 2017-01-06
Default components entity

dmarcos:
The list of default components should not be hard coded. We should get it from each entity since they can change. The `a-scene` default components are different than those from `a-entity`


dmarcos:
This will fix the problem with `a-scene` in https://github.com/aframevr/aframe-inspector/issues/409

fernandojsg:
Cool, so it's secure to change the rest of the code where we're using currently DEFAULT_COMPONENTS right?

dmarcos:
I think so

PR #420, opened 2017-01-10
Fixes #417 ExitVR when opening the inspector

### LINKED ISSUE #417: Inspector shown in stereo when entering while in VR mode
When entering the inspector while still in VR mode the inspector shows a stereo (L/R camera) view. This is not helpful as the inspector is used on the desktop. Workaround is to exit VR mode before entering inspector, but sometimes I want to "freeze" the layout immediately from my VR device orientation which makes it difficult to exit VR prior to opening inspector. Screenshot:
screenshot 12

fernandojsg:


fernandojsg:
tested both in fullscreen and VR

Issue #402, opened 2016-11-28
Broken styles


ngokevin:
Working on preparing slides for a meetup. I found the latest Inspector has broken styles. The CSS canvas is not covering the height and width.

I also find clicking objects doesn't work well. I often click on an object, and nothing responds, and I have to fidget with the camera to get it to be selected. (last working at 07550209aee1441f102b4dece6d7464e1867b66f)

![1](https://cloud.githubusercontent.com/assets/674727/20672230/5b0be3f6-b535-11e6-8b38-42d8d9e952d0.PNG)

fernandojsg:
I've found the same issue on some windows machines, it seems to be related to https://github.com/aframevr/aframe-inspector/blob/master/src/css/main.css#L777
I believe we should use `zoom: 1` instead. Could you give it a try in your slides?

ngokevin:
Yeah, we found this a while back https://github.com/aframevr/aframe-inspector/blob/master/example/embedded-zoom.html

ngokevin:
Nevermind, different zoom issue.

fernandojsg:
Yes, with the `embedded-zoom` example I believed that `zoom: reset` should works, but now it doesn't seems to work on highres windows. In fact I've found that using 50% zoom on these machines shows the canvas correctly, I don't know why. I should investigate a bit more

kfarr:
I still see this issue projects with newest aframe master, for example here: 
![screenshot 12](https://cloud.githubusercontent.com/assets/470477/21704480/9da5b8b6-d36e-11e6-87cc-0762decf2785.png)

fernandojsg:
@ngokevin @kfarr could you verify if it's already working correctly with the changes in https://github.com/aframevr/aframe-inspector/pull/421 ?

kfarr:
I am still seeing the attached, this is using A-Frame Version: 0.4.0 (Date 09-02-2017, Commit #27713c2) on this url http://aframe-city-firebase.surge.sh/nofirebase.html
![screenshot 45](https://cloud.githubusercontent.com/assets/470477/22773905/70a970cc-ee59-11e6-9f5b-1ea7a99ab571.png) on firefox nightly and chrome (mainstream release)

kfarr:
![screenshot 8](https://cloud.githubusercontent.com/assets/470477/23294419/3b045e82-fa20-11e6-8e9e-443e678e7c51.png)
Chromium latest webvr build (Feb 17) working!

[...]

PR #421, opened 2017-01-10
Set zoom:1 to canvas

fernandojsg:
It should fix the issues (https://github.com/aframevr/aframe-inspector/issues/402) with chrome zoom on HDPI Windows (`devicePixelRatio=1`).
I've tried the default example and the `embedded.html` on a windows with `pixelRatio=2` and `pixelRatio=1` and it's working.
Could you give it a try @ngokevin @kfarr @nucliweb ?

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -774,7 +774,7 @@ a.help-link:hover {
   background-color: #191919;
   position: fixed;
   z-index: 9999;
-  zoom: reset;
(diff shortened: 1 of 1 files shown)


dmarcos:
Looks great! It fixes the issue on my surface book

dmarcos:
Before before

After after

PR #422, opened 2017-01-11
Added version and build timestamp

fernandojsg:


webpack.config.js
```diff
@@ -13,13 +13,31 @@ if (process.env.NODE_ENV === 'dev') {
...
+
 // Minification.
 var plugins = [
   new webpack.DefinePlugin({
     'process.env':{
       'NODE_ENV': JSON.stringify(process.env.NODE_ENV)
-    }
-  })
+    },
+    VERSION: JSON.stringify(require('./package.json').version),
```
dmarcos, line 37: do you need `stringify`?

webpack.config.js
```diff
@@ -13,13 +13,31 @@ if (process.env.NODE_ENV === 'dev') {
...
+    BUILD_TIMESTAMP: getBuildTimestamp()
```
dmarcos: Is it better to use the git hash instead of the timestamp?

dmarcos (commented):
It looks good

webpack.config.js
fernandojsg: Yeah you're right, not needed, I just copied it from the previous plugin O:)

fernandojsg:
Updated to show hash and a more readable timestamp:
```
A-Frame Inspector Version: 0.4.0 (11-01-2017 Commit: b53002f)
```

dmarcos:
It looks great. We should do the same for a-frame

PR #424, opened 2017-01-25
Added Keyboard events to modify the values of rotation, translation and scaling using keyboard

vershwal:
It is a common shortcut in softwares like blender to change the transform properties in the three dimensions using the keyboard.
This P.R. implements the same.

If you press 'r', you go to the scaling mode. Now pressing 'x' followed by a numerical value modifies the scale in the x direction to that value on pressing enter.

Similarly for other transform properties and dimensions.

### DIFF
--- a/src/lib/shortcuts.js
+++ b/src/lib/shortcuts.js
@@ -2,6 +2,16 @@
 var Events = require('./Events');
 import {removeSelectedEntity, cloneSelectedEntity} from '../actions/entity';
 
+AFRAME.Keyevents = {};
+AFRAME.Keyevents.VAL = null;
+AFRAME.Keyevents.DIM = 'x';
+AFRAME.Keyevents.LASTKEY = null;
+AFRAME.Keyevents.NEGATION = false;
+
+document.addEventListener("click", function(){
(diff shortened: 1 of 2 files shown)


vershwal:
Hi @fernandojsg 

I have implemented a shortcut feature for aframe-inspector.
Could you review the P.R.?

fernandojsg:
While it's a nice feature I think it's a bit overkill feature right now for the current target of the inspector, probably if in the future the editor would evolve in something more like a content creation and composition tool instead of a devtool it could makes more sense.
What do you think @feiss ?

feiss:
Yeah, I agree. It is a nice feature and well implemented, but it is quite hardcore for the current status of the inspector, it is not prepared for this interaction level (yet).

It will be interesting to reopen this PR in the future, for sure

fernandojsg:
Thanks. Ok so let's close it by now and we'll reopen it in the future if we seems to evolve in that way.

vershwal:
Thanks for the review.
Glad you guys liked it.

Yes I have to agree this P.R. might not be a very valuable addition to the inspector. 
Its more of a compositor feature.

PR #426, opened 2017-02-06
Added functionality to copy and paste entities

vershwal:


feiss:
Very nice, thank you!

I've found a little bug in this: if you have an input box (like 'search', 'add component', or 'mergeTo') focused, both copy and paste actions are performed, and they should not. Could you fix this?

vershwal:
Hi @feiss ... Thanks for pointing that out. 
I've fixed the bug using document.activeElement.

You can check it here : 16ebe915510a338a4aee8adb001e0ae854d509ea

fernandojsg:
Hi @vershwal thanks for the PR.
The main concerns that comes to my mind is how to detect when to use Ctrl and when to use Command, as on OSX you copy with CMD+C instead of CTRL+C. Any ideas how to handle that?

vershwal:
Hi @fernandojsg .. Thanks for pointing that out.

So it turns out there exist metaKey and ctrlKey attributes of the event parameter received on a keyPress.

These return true or false depending on weather or not command (metaKey) or control(ctrlKey) key is  pressed. This made the fix simpler.

Modified the previous approach to include this over here : https://github.com/aframevr/aframe-inspector/pull/426/commits/0e640ab32b271d3b191ccbf0c2cd7b373fed6412

fernandojsg:
Thanks @vershwal ! :)

Issue #175, opened 2016-07-08
Extend the scenegraph's filter to components and attributes


fernandojsg:
Currently the filter in the scenegraph is filtering tagName and ID of each entity, it could be great to let it filter through the component's name and event values. For example typing `light` it will filter the entities that has a ID `light` but also the ones that have a `light` component. Or even typing `box` could filter the geometries that has a `box` attribute set.

PR #431, opened 2017-02-08
Extended filter

fernandojsg:
Fixes https://github.com/aframevr/aframe-inspector/issues/175
Added extended scenegraph filter:
**Before**
- ID
- tagName

**Now**
- Components names
- Component's values

I was about to add the component's attributes' names but I believe it's not that useful.

extfilter


feiss:
Really cool. I don't think there's need for component's attribute names, too.

PR #434, opened 2017-02-09
more subtle grid color

feiss:


fernandojsg:
I'm having a hard time to differentiate the colors from background on the macbook :D maybe a little less subtle? X)

feiss:
changed. What about now?

Before:
image

After:
image

fernandojsg:
Thanks!

PR #436, opened 2017-02-09
CMD/CTRL+P to focus filter box

fernandojsg:


feiss:
works nice!

why ctrl+p? it is a shortcut used for printing usually.. could ctrl+f be overwritten? (`f` for filter or find, default when searching)

also, I'd expect `esc` key to work inside the filter box. It would call the same function as the clear filter button

fernandojsg:
I used `ctrl+p` as is commonly used on sublimetext/atom/devtools to jump to another file so people got used to that, but I'm open to other suggestions yep <= @ngokevin @dmarcos what do you think?

Regarding `esc` it sound cool for me yep I'll do that too.

fernandojsg:
@feiss ESC implemented

fernandojsg:
I've just checked and `ctrl|cmd + f` can be overwritten, so probably it could be a better idea to use it for the filter, and leave the `ctrl|cmd + p` for later use like in the case we'll include files to edit or so...

Issue #433, opened 2017-02-09
saveSceneToHTML() exports with an inactive camera, making an empty looking scene


Utopiah:
It's possible to fix it starting and closing the inspector or making the camera active `document.getElementById('camera').setAttribute("camera", "active:true")` but it would be more natural to change the camera to active on export (either on `saveSceneToHTML()` or even `generateHtml()`.

fernandojsg:
@Utopiah that should be fixed with the latest changes, do you mind giving it a try?

Utopiah:
Tried using `master` and it works, namely the scene is visible but it's not using the existing camera as it is still marked `active:false` thus injects a new camera. Not sure if that's the intended behavior.

fernandojsg:
Ah ok I see. I was just deleting the default camera injected by the editor, but you're right that I need to remove the `active:false` on the default camera.

ngokevin:
https://github.com/aframevr/aframe-inspector/issues/517#issuecomment-445454902

PR #440, opened 2017-02-10
fixing the active camera on saving

Utopiah:
Currently the active camera is the editor yet, once saved, the editor is not opened hence showing a grey page without content (cf #433 ). This PR switches the active camera from editor camera to main camera.

### DIFF
--- a/src/components/scenegraph/Toolbar.js
+++ b/src/components/scenegraph/Toolbar.js
@@ -28,6 +28,8 @@ export default class Toolbar extends React.Component {
       // URL.revokeObjectURL(url); breaks Firefox...
     }
     function saveString (text, filename) {
+      text = text.replace("active:true", "active:false")
+      text = text.replace('camera="active:false"', 'camera="active:true"')
       save(new Blob([ text ], { type: 'text/html' }), filename);
     }
     var sceneName = getSceneName(document.querySelector('a-scene'));


dmarcos:
Should we change the active camera to the scene one and back to the editor camera when saving? Maybe better than replace strings. I

fernandojsg:
Yep, I agree with @dmarcos , I wouldn't rely on string replacements, for example let's say that the exporter will modify in a near future to `camera: false` adding an extra space, or what happens if we've more than one camera, this will activate all of them. 
It's better to switch the previous active camera in the scene, export and then go back to the editor's camera.
Closing it, if you feel like to modify to fix in that way, we'll just reopen it! :)

Utopiah:
Wouldn't that make a weird user experience to see a camera change as he/she saves?

dmarcos:
if you change, save and change back the user won't perceive anything since everything will happen before the render loop

Utopiah:
OK let me try that. It's the alternative I already had in mind but wasn't sure which one was "cleaner".

Utopiah:
@fernandojsg what about 
```
    INSPECTOR.close();
    var sceneName = getSceneName(document.querySelector('a-scene'));
    saveString(generateHtml(), sceneName);
    INSPECTOR.open();
```
instead?

Gives the expected outcome and I didn't notice any UX issue.

PR #445, opened 2017-02-11
Fixes #425

### LINKED ISSUE #425: Outline of object remains even after deleting it
I came across this issue while browsing.
In the example page, if we delete the biggest box, its outline seems to remain there and is clickable.

screenshot from 2017-02-06 16-31-48
screenshot from 2017-02-06 16-32-02
screenshot from 2017-02-06 16-32-22
screenshot from 2017-02-06 16-32-33

This doesn't happen for other objects though.

vershwal:


fernandojsg:
@vershwal nice catch! thanks! :)

PR #463, opened 2017-03-31
Use a delimiter that isn't valid in NPM package names.

donmccurdy:
The '.' character used by libraries like Lodash that publish sub-packages to NPM. I've taken the same pattern with `aframe-extras._____`, which is breaking here. [Spaces are invalid in NPM package names](https://github.com/npm/validate-npm-package-name#naming-rules) as well as component names, hence the choice.


fernandojsg:
Thanks!

PR #470, opened 2017-04-20
Selection should select the object closest to the camera.

andyli:
`raycaster.intersectObjects(objects)` returns intersections sorted by distance, closest first, according to the doc at https://threejs.org/docs/?q=Raycaster#Reference/Core/Raycaster

So, once we find a valid selection target, we should break the loop, because the later ones will be further away from the camera.

### DIFF
--- a/src/lib/viewport/index.js
+++ b/src/lib/viewport/index.js
@@ -210,6 +210,8 @@ function Viewport (inspector) {
           } else {
             inspector.selectEntity(object.el);
           }
+
+          break;
         }
 
         if (!selected) {


andyli:
A demo of the bug and fix:

The original A-Frame hello world:
https://aframe.io/examples/showcase/helloworld/
Put up the inspector, try click on the cylinder and the floor behind it will be selected.

The same demo with the fixed inspector:
https://aframe-inspector-selection-fix.glitch.me
Put up the inspector. The cylinder can be selected by clicking on it.

fernandojsg:
@andyli sorry for the late merge. Good catch! thanks

PR #485, opened 2017-06-13
Fixes #443 setting active:true to scene camera

fernandojsg:
/cc @Utopiah

### DIFF
--- a/src/lib/exporter.js
+++ b/src/lib/exporter.js
@@ -1,4 +1,5 @@
 import {getClipboardRepresentation} from '../actions/entity';
+const INSPECTOR = require('../lib/inspector.js');
 
 /**
  * Get scene name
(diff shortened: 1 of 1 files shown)


Utopiah:
As is I get `<a-entity id="camera">` but not `<a-entity id="camera" camera>` so a camera still get injected

fernandojsg:
It was driving me nuts but now it seems it's a bug from the `clipboardRepresentation` of these entities: https://github.com/aframevr/aframe-inspector/issues/486

Utopiah:
Not surprised there were quite a few discussion on that "cleaning" up part, it's not trivial :/

fernandojsg:
Yep, 100% cleaning it won't be possible unless we'll start a scene from scratch and we control everything that is injected there, as the user could have plenty of dynamic code around and we can't just handle all of that, but at least we should delete everything that we inject (both the inspector and aframe itself)

fernandojsg:
@Utopiah do you mind giving it a try again? It should work with the latest changes from @delapuente

Utopiah:
No entities are visible anymore. I believe it comes from saving position to `position="x:0;y:8;z:0"` instead of `position="0 8 0"`.

fernandojsg:
Ouch I didn't notice that -_- /cc @delapuente I guess that's because of the last PR https://github.com/aframevr/aframe-inspector/pull/488

fernandojsg:
@Utopiah hopefully one last try? :)

We've a known issue because of how the serialization is done by aframe that will need some extra postprocessing but I'll leave that for another PR. Basically if you reference a asset for example, it will just export it using the url but not the ID:
```
 <a-entity geometry="" material="src:https://aframe.io/sample-assets/assets/images/wood/crate.gif" position="3 4 0"></a-entity>

<a-entity geometry="" material="src: #crateImg" position="3 4 0"></a-entity>
```

Utopiah:
Works!

fernandojsg:
Finally! thank you so much for the help @Utopiah and of course @delapuente 💋

PR #487, opened 2017-06-14
Telling inherited components apart form entity-owned components.

### LINKED ISSUE #486: Component with default attributes deleted from the clipboard representation
If an entity has a component without any parameter, or with the default parameters it won't be included on the clipboard representation, for example `<a-entity geometry=""></a-entity>, <a-entity camera="active: true"></a-entity>...`

Steps to reproduce:
- Add a new entity
- Add a `geometry` component
- Copy to clipboard and you'll get `<a-entity></a-entity>`
- Modify any parameter from it, for example depth, and copy again: `<a-entity geometry="depth:1.85"></a-entity>`

It's also happening in https://github.com/aframevr/aframe-inspector/pull/485 as if you have a camera without any parameter it will delete the component from the clipboard representation too.

/cc @delapuente

delapuente:
Fix #486

The optimizeComponents() routine had a path in which the component was removed
if the optimal update was `null`, pointing that there were no difference between
the implicit value and the current value. But this assumption made components
with only default attributes (as in <a-entity geometry=""></a-entity>) to be
removed.

The modification makes getImpliciValue() to indicate if the component is
inherited from injected components or mixins, or if it is genuinely owned by the
entity. Just in the case the component is inherited and it does not need an
update, the component can be safely omitted.

### DIFF
--- a/src/actions/entity.js
+++ b/src/actions/entity.js
@@ -180,10 +180,13 @@ function optimizeComponents(copy, source) {
   var components = source.components || {};
   Object.keys(components).forEach(function (name) {
     var component = components[name];
-    var implicitValue = getImplicitValue(component, source);
+    var result = getImplicitValue(component, source);
+    var isInherited = result[1];
+    var implicitValue = result[0];
     var currentValue = source.getAttribute(name);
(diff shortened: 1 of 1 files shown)


fernandojsg:
Cool, thanks! ;)

PR #488, opened 2017-06-14
Avoiding setAttribute() side effects.

delapuente:
While optimizing components for the clipboard representation, calling
setAttribute() directly could have side effects on entity's components already
processed (such as re-setting properties already removed). To avoid side
effects, HTMLElement original setAttribute() and removeAttribute() functions are
used instead.


fernandojsg:
Perfect, thanks! ;)

fernandojsg:
@delapuente I've just reverted it because of the issue related to the position format that it generates

PR #500, opened 2017-07-18
update editorcontrols to fix firefox mousewheel

ngokevin:


fernandojsg:
@ngokevin just out of curiosity, is this the updated version of https://github.com/mrdoob/three.js/blob/dev/examples/js/controls/EditorControls.js ? or did you add/remove something extra?

ngokevin:
Yeah from three

fernandojsg:
Cool! thanks ^_^

PR #526, opened 2017-10-25
Use PropTypes from its own package and fix all lint errors

michaltakac:


fernandojsg:
Thank you very much @michaltakac !

PR #529, opened 2018-01-06
Implemented collapsible elements in scenegraph

Adam4lexander:
Implemented similarly to Unity. Click carrots to expand/collapse elements.
While searching all collapsed elements are expanded. When clearing the search the selected element will have its parent elements expanded.
I used a weakmap to store expanded/collapsed state on elements so there was no need to mess with the rebuildOptions code.

### DIFF
--- a/src/components/scenegraph/SceneGraph.js
+++ b/src/components/scenegraph/SceneGraph.js
@@ -39,7 +39,8 @@ export default class SceneGraph extends React.Component {
       value: this.props.value || '',
       options: [],
       selectedIndex: -1,
-      filterText: ''
+      filterText: '',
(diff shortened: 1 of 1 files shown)


ngokevin:
Thanks! I'll try to get around to looking at this. Very useful.

ngokevin:
Thanks!

ngokevin:
This might've affected search, I don't think it's working for me.

ngokevin:
Does search work for you? Doesn't work well for me, not sure if it was broken before already.

Adam4lexander:
It's not working for me either. There's a good chance I broke it since I was touching that code. Sorry about that I'll take a look and make a fix.

PR #530, opened 2018-01-07
Press 'f' to focus on currently selected entity

Adam4lexander:
I find this feature really useful working on complex scenes. Most 3d programs map the f key to this function so it was bugging me without it. Added the f hotkey which will focus on the selected element. Automatically zooms in or out to frame the selection in the viewport.

I had to edit the THREE.EditorControls file because it didnt support auto zoom and it had a bug in its focus function anyway. Should I try and make a merge request with the Three repo too?

I want the camera to lerp to the target focus point too, I'll add this next if this change is accepted.


ngokevin:
It still enters full scene for me. Maybe shift/f?

ngokevin:
Would consider doing double-click on entity (in the scenegraph or viewport) to focus

ngokevin:
Also yeah would be good to PR those EditorControls changes to three.js.

Adam4lexander:
Sounds good to me, I'll rebind to shift/f and also map double clicking to focus. I'll PR into three.js first to make sure we're in sync.

ngokevin:
Thanks! Actually `shift/f` would probably still trigger it, I think we can keep `f`, maybe we disable the keyboard shortcuts as you did before? I don't know if we want to allow the fullscreen/exitVR in Inspector anyways.

Adam4lexander:
Ok then, I agree with you there, since entering vr mode whilst the inspector is open doesn't give very good results. I'll go ahead and disable it in inspector mode.

ngokevin:
Sounds good!

Adam4lexander:
Finally got this change into ThreeJS r91. Going to update here soon.

Adam4lexander:
Ok done. The new EditorControls has been copied from threejs. Pressing 'f' will focus on selected entity. Double clicking the entity tag in the scene graph will also focus on it.

When the inspector starts it disables the keyboard-shortcuts component to stop it entering vr mode when 'f' key is pressed. Although maybe it would be better if the keyboard-shortcuts component stopped listening for key presses when paused?

ngokevin:
Ah, that would be smart. Let's do that.

Adam4lexander:
Done I've taken out the code which disables keyboard-shortcuts here and made a separate pull request in aframe: https://github.com/aframevr/aframe/pull/3413

ngokevin:
Thanks!

PR #531, opened 2018-01-26
fix search in scenegraph

Adam4lexander:
Looks like I made a bug! Just had to re-arrange a couple of lines and its good.


ngokevin:
Thanks!

Issue #532, opened 2018-01-27
Improve keyboard navigation of collapsed entity tree


ngokevin:
Right now, hitting Down will enter the tree but not expand it.

I would skip to the next non-collapsed entity on Up or Down.

And have Left or Right collapse/uncollapse the tree

@Adam4lexander

Adam4lexander:
np Im on it.

PR #533, opened 2018-01-28
Improve scenegraph keyboard navigation

Adam4lexander:
#532 

Fixed the keyboard controls as you suggested.

Fixed another issue where a user selects an object in the viewport which has collapsed parents in the scenegraph. Now those parents are forced to expand to make the selection visible in the scenegraph.


ngokevin:
Thanks again!

PR #537, opened 2018-02-25
Remove aframe registry

fernandojsg:
Removing the aframe-registry component loader based on the discussion https://github.com/aframevr/aframe-registry/issues/78.

I've just commented the code right now, as many people are complaining about the error, but if in the near future we decide to give it another opportunity we could just uncomment it.
If after a while we see that it doesn't worth the effort we could just remove the whole logic from it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12393 @@
+{
+  "name": "aframe-inspector",
+  "version": "0.7.7",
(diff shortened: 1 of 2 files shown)


ngokevin:
Sweet. I'd vote to remove the code. We winded down the Registry (later we can replace it with a static website that shows npm results plus some essential ones).

PR #540, opened 2018-03-11
remove sidebar width

arpu:
with this i can see all entity
bildschirmfoto von 2018-03-11 02-10-00

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -218,7 +218,6 @@ input.error {
   overflow: auto;
   position: fixed;
   top: 0;
-  width: 230px;
 }
 
 #sidebar {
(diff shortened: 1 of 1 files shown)


ngokevin:
Hmm on lower res screens, this would make the entity panel cover the scene. Sure there's a way to compromise

arpu:
this is how its looks now on ipad 
so adding x scrolling on the sidebar ? 

bildschirmfoto von 2018-04-09 21-04-10

ngokevin:
Does the sidebar not scroll already? If it doesn't, yeah, let's add x scrolling.

arpu:
no it does not scroll :/ i have no idea how i can add the scrolling :/

arpu:
will open a new pull request with x scrolling

Issue #552, opened 2018-07-26
Local example not working


donmccurdy:
After following the instructions for running locally, there is an error from `inspector.js` in the onDomLoaded callback:

```js
    this.sceneEl = AFRAME.scenes[0];
    if (this.sceneEl.hasLoaded) {
      this.onSceneLoaded();
    } else {
      this.sceneEl.addEventListener('loaded', this.onSceneLoaded.bind(this));
    }
```

`AFRAME.scenes` is empty at this point, so the second line throws an error.

Issue #523, opened 2017-10-17
Consider supporting user inspector plugins?


donmccurdy:
Hi @fernandojsg —

What do you think about letting developers write plugins for the inspector, by having a simple API for attaching their DOM element to a tab?

Some example use cases:

- Export current scene to glTF or OBJ
- Search for models from different model libraries
- Custom animation timeline editor, similar to [frame.js](https://github.com/mrdoob/frame.js/)
- Creating and exporting a navmesh derived from the current scene
- More detailed performance panel, explaining how many entities you have, number of materials, etc.

These all seem like interesting use cases for the inspector, but not necessarily something we'd build and support ourselves. But maybe we could have an API like this?

```js
var el = document.createElement('div');
var btnEl = document.createElement('button');
btnEl.textContent = 'Export Scene';
btnEl.addEventListener('click', () => ...);
el.appendChild(btnEl);

AFRAME.inspector.registerPanel(el, {position: 'bottom'});
```

dmarcos:
In order for plugins to not be disruptive to the current UI I would probably give them their own panel to render themselves. That panel would be invoked by the user. Idea:
- A `plugins` button above the `scene graph` view (next to the motion capture, save... buttons) that replaces the `scene graph` view with that of a selected plugin. The drawback would be you could not see multiple plugins or the scene graph and a plugin simultaneously.

donmccurdy:
Yeah, I think the location is not too critical unless it becomes used for a bunch of fancy stuff. Which would be a good problem to have. Could be a third, new panel that pops up from the bottom of the page as well.

fernandojsg:
Actually I started with something called `modules` a while back when I wanted to create a VR editor mode for the inspector. Basically you register the modules and they can be activated or deactivated. As I didn't continue with the VR editor I've just the `dummy` module there, but we could use it just as a starting point for something like you propose.
https://github.com/aframevr/aframe-inspector/blob/d3e0a32dfe0a9100386c1e6ec561f3003c48d526/src/lib/modules/dummy/index.js

fernandojsg:
@donmccurdy @dmarcos do you have any idea that could be simple enough to use it as an example and as a proof of concept of the plugin system?

Utopiah:
- @tomas-polach from the Archilogic3D/3D.io team already [customized the inspector for Google Blocks integration](https://twitter.com/TomasPolach/status/915677320429621254) 
[...]

PR #553, opened 2018-07-26
Export to .glb with embedded images.

### LINKED ISSUE #547: glTF export uses absolute URLs to images
I think it would be preferable to always embed images, or at least give the option. When absolute image URLs are used, I can't import the resulting `.gltf` into Blender and validation gives errors.

### LINKED ISSUE #551: Aframe inspector: glTF export button broken/ not working as expected
Trying to export my aframe scene using this button was not initiating a download at all,

Instead I selected the entity and exporting that as a glTF using the button on the right and it would download

attempting to re-display the model in aframe was causing issue and I was advised to try viewing it on

https://gltf-viewer.donmccurdy.com/
to which the response was missing texture:

    A-Frame Version: 0.8.2
    Platform / Device: OSX chrome

donmccurdy:
@fernandojsg  I believe this PR fixes #547 and fixes #551, but due to #552 I am not sure how to test this locally.

### DIFF
--- a/src/components/components/CommonComponents.js
+++ b/src/components/components/CommonComponents.js
@@ -8,7 +8,7 @@ import Mixins from './Mixins';
 import {updateEntity, getClipboardRepresentation} from '../../actions/entity';
 import Events from '../../lib/Events';
 import Clipboard from 'clipboard';
-import {saveString} from '../../lib/utils';
+import {saveBlob} from '../../lib/utils';
 
 // @todo Take this out and use updateEntity?
 function changeId (componentName, value) {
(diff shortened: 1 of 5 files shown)


fernandojsg:
@donmccurdy after a quick pass I've just tried it and it "works" but the .glb file contains just `[object Object]`. I'll check the code to see what is going wrong.

But I don't get any error because of https://github.com/aframevr/aframe-inspector/issues/552 I can't reproduce it /cc @dmarcos @ngokevin any idea what could be causing it?

fernandojsg:
@donmccurdy I believe one of the needed steps is to update `GLTFExporter.js` with an updated version from three.js that supports `binary: true` :D

donmccurdy:
Ok, I think this is working now — updated GLTFExporter, marked inspector helpers (see https://github.com/aframevr/aframe-inspector/issues/523#issuecomment-384522834), and filtered them out while exporting.

arpu:
i had some time to test this! and it works fine

fernandojsg:
Yeah, thanks for remind me I forgot about this one! Thanks @donmccurdy and @arpu

arpu:
@donmccurdy  if i use 
    map.wrapS = THREE.RepeatWrapping;
     map.wrapT = THREE.RepeatWrapping;
    map.repeat.x = 20;
     map.repeat.y = 20;
and export the generated gltf model this is not applied to the exported model 
is this a known problem ?

donmccurdy:
Texture repeat, rotation, and offset are not part of the core glTF spec, but require a very recent extension. The THREE.GLTFLoader implementation has not been merged yet (https://github.com/mrdoob/three.js/pull/13831), and no PR for THREE.GLTFExporter exists yet.

EDIT: The wrapS / wrapT part should already be supported, though: https://github.com/mrdoob/three.js/blob/dev/examples/js/exporters/GLTFExporter.js#L740-L741

Issue #578, opened 2019-02-26
clear filter button does not work


feiss:
after filtering the scenegraph, if you click on the clear filter button it doesn't go back to the full scene list.

vincentfretin:
It works on latest version. @dmarcos you can close.

PR #580, opened 2019-03-08
Fix: clear filter button does not work #578

aj019:
The clear filter button does not work in inspector
Fixed the following

Cleared filter input box
Update filtered entity list


ngokevin:
Thanks!

PR #581, opened 2019-03-09
fix prettier eslint

lazychip:
When i cloned the repo and started to look the code eslint is so messed up and all those end lines with my eslint on 😖.  
I have added prettier-eslint to fix the code and husky hooks on precommit to future save the code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "aframe-inspector",
-  "version": "0.8.14",
+  "version": "0.9.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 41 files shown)


ngokevin:
I just have the codebase on Prettier. I don't think it matters much to use Prettier vs. ESLint, so it's okay to just keep it on Prettier? More common in React/JSX codebases.

lazychip:
Personally i like eslint and I think eslint, prettier codebase is more common in React/JSX. Its okay to keep the codebase on Prettier but i think if you want to continue the codebase on prettier only  then remove the eslint config from the codebase to avoid the confusion for the contributors.

vincentfretin:
@dmarcos Please close this one. All eslint errors are fixed by #639 and the prettier reformat will be done in #647

dmarcos:
Close in favor of https://github.com/aframevr/aframe-inspector/pull/647

PR #585, opened 2019-08-09
fix when saving the scene with a-frame watcher and color is changed

khcd:
I noticed when changing material color it wasn't being saved with aframe watcher due to an undefined in the `entityupdate` listener,  other material properties seemed to be ok. I didn't test all of them as color was main concern. 

Let me know if this is the best way to fix it.

previous payload when changing color:

`{"sphere1":{"position":"0 1.234885180247932 -5","material":{},"box1":{"material":{}}}`

now:

`{"sphere1":{"position":"0 1.234885180247932 -5","material":{"color":"#4494ef"}},"box1":{"material":{"color":"#d93336"}}}`

**Additions: fix CI build, I choose node 8 to be safe although can be changed to LTS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - 4
+  - 8
 
 install:
   - npm install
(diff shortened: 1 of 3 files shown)


khcd:
@fernandojsg Any feedback on this one? I can create a corresponding issue if required.

PR #624, opened 2021-09-23
Update gltf exporter

kfarr:
- gltf exporter was outdated and did not support export of things such as repeated textures
- ~~transform controls used legacy three.js syntax that raised repeated console warnings~~ (removed from this pull request as it was resolved via https://github.com/aframevr/aframe-inspector/pull/627)


kfarr:
removed transform controls from this pull request, @dmarcos could use your review

dmarcos:
Thanks!

Issue #625, opened 2021-09-24
Transform controls error blocks inspector loading with A-Frame master due to three.js breaking changes


kfarr:
**Description:**

- A-Frame Version: Master (https://cdn.jsdelivr.net/gh/aframevr/aframe@d52af46565230a33c0fa23e045fb74e877df7dc9/dist/aframe-master.min.js)
- Platform / Device: Desktop Mac OS
- Reproducible Code Snippet or URL: 

Working in A-Frame 1.2: https://aframe.glitch.me/ and press ctl-alt-i
Not working in A-Frame master:
https://inspector-broken-master.glitch.me/ and press ctl-alt-i

No inspector loads, and there is a console error:
`Uncaught TypeError: Class constructor MeshBasicMaterial cannot be invoked without 'new'`

**Cause:**
This appears to be related to a breaking change in an updated version of three.js relating to the TransformControls used by the inspector. 

**Solution:**
A potential solution is to update the TransformControls in inspector to a newer version.

kfarr:
I found the latest version of es5 transform controls here:
https://github.com/mrdoob/three.js/blob/master/examples/js/controls/TransformControls.js

I attempted to copy the entire contents of that file into this repo's src/lib/TransformControls.js file.

The inspector now loads, which is an improvement, but there is now a new issue:
- The transform gizmo does not appear as expected when an object is selected. There is an error on the console: `TransformControls.js?0d71:12 Uncaught TypeError: Class constructor Object3D cannot be invoked without 'new'`

After some google searching this appears to be related to trying to use a module intended for ES6 from an ES5 environment. I am not sure how to proceed. Next I am looking for older versions of TransformControls prior to ES6 conversion that might still work with this latest version of three.js.

kfarr:
This may be an issue with the way that the inspector's viewport.js file imports and attempts to create a new transformControls object

Here is the legacy ES5 import: 
https://github.com/aframevr/aframe-inspector/blob/master/src/lib/viewport.js#L5

And then here is the first invocation:
https://github.com/aframevr/aframe-inspector/blob/master/src/lib/viewport.js#L51

donmccurdy:
[...]

PR #629, opened 2021-10-29
Add 'terser-webpack-plugin-legacy'

### LINKED ISSUE #628: npm install raises error
There is a console error when running `npm install` on master repo after PR https://github.com/aframevr/aframe-inspector/pull/627
```
ERROR in aframe-inspector.min.js from UglifyJs
Unexpected token: name (NumberWidget) [aframe-inspector.min.js:3194,6]
```

To get npm install to work without this error, in .babelrc line 3 change `exclude` to `include`, then run `npm install` again and it appears to install correctly.

However, then you must to change it back to `exclude` in order for the build to run and for the inspector to correctly display on localhost when running `npm start`

donmccurdy:
Fixes #628, #625.

Fixes minified builds; unminified builds were already working.

This PR is intended to be the minimum set of dependency updates needed to get the Inspector working with ES6 Classes, following up from #627. The version of UglifyJS bundled with older Webpack versions didn't support ES6, so I'm bumping to Webpack 3 and installing [Terser](https://www.npmjs.com/package/terser-webpack-plugin-legacy). Minified bundle size decreases from 440kb to 410kb.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,14551 @@
 {
   "name": "aframe-inspector",
-  "version": "0.8.14",
-  "lockfileVersion": 1,
+  "version": "1.2.0",
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "aframe-inspector",
+      "version": "1.2.0",
+      "license": "MIT",
+      "dependencies": {
+        "autoprefixer": "^9.4.2",
+        "babel-core": "^6.26.3",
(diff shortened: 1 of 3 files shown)


donmccurdy:
@dmarcos just FYI, I think this PR is a necessary part of #627.

kfarr:
Getting support requests on this from users not able to build. Any issues to resolve before ok to merge @dmarcos ?

dmarcos:
Thank you for the reminder. I missed this.

PR #637, opened 2022-10-02
Adjustments for THREE r144

vincentfretin:
Some fixes to work with aframe master (tested against three r141 and not r144 because there is no updated build currently on aframe repo) and fixing some issues I found while testing, see each commit for details.

PR #638, opened 2022-10-04
Fix some UI issues

vincentfretin:
Needs to be rebased after #637 is merged, the branch won't work if you're trying to test it

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -126,7 +126,10 @@ Inspector.prototype = {
       helper.visible = false;
       this.sceneHelpers.add(helper);
       this.helpers[object.uuid] = helper;
-      helper.update();
+      // SkeletonHelper doesn't have an update method
+      if (helper.update) {
+        helper.update();
+      }
     };
   })(),
 
--- a/src/index.js
+++ b/src/index.js
@@ -103,12 +103,6 @@ Inspector.prototype = {
   },
 
   addHelper: (function() {
-    const geometry = new THREE.SphereBufferGeometry(2, 4, 2);
-    const material = new THREE.MeshBasicMaterial({
-      color: 0xff0000,
-      visible: false
-    });
-
     return function(object) {
       let helper;
 
--- a/src/components/components/CommonComponents.js
+++ b/src/components/components/CommonComponents.js
@@ -105,6 +105,7 @@ export default class CommonComponents extends React.Component {
           className="gltfIcon"
           onClick={event => {
             this.exportToGLTF();
+            event.preventDefault();
             event.stopPropagation();
           }} >
           <img src={process.env.NODE_ENV === 'production' ? 'https://aframe.io/aframe-inspector/assets/gltf.svg' : '../assets/gltf.svg'} />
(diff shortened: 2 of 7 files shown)


dmarcos:
I think it doesn't need rebase. does it?

vincentfretin:
If you want to merge it right away, yeah it doesn't need a rebase.

vincentfretin:
I rebased in case you want to test the branch first before merging.

vincentfretin:
Once this is merged, please run `npm run dist` and commit the build so we can use a jsdelivr gh url to on the aframe inspector component.

vincentfretin:
I have a `fix-some-ui-issues-with-dist` branch that is based on this PR where I committed the dist. There is also a commit to fix some style issues when the inspector open on a page that includes bootstrap or tailwindcss reset styles. I'll do a separate PR for this, there are still 2-3 things that needs to be fixed.
You can test in your project with
```
<a-scene inspector="url:https://cdn.jsdelivr.net/gh/vincentfretin/aframe-inspector@a6ab6b985138c12a9094a13cf47fb6173e18a737/dist/aframe-inspector.js">
```

vincentfretin:
I'm currently testing with aframe master threejs r144 with a tailwindcss theme. The latest commit I did fixes the error update method undefined because I had an avatar with SkinnedMesh in my scene.

vincentfretin:
I did some cleanup in the files in the lib folder because I'm reusing all those files but with my own ui in a project, so I went ahead and contributed it here too. I think you can merge now, those are really small changes.

Are you interested in more heavy updates, update the webpack stack, run prettier on the files again, remove the eslint rules concerning the style, integrate husky to run prettier on the files, update to latest react and api, use only import/export everywhere and not a mix of import and require. Will you merge it?

vincentfretin:
I moved the code cleanup I'm currently doing in #639 so you can concentrate on the functional issues I fixed in this PR. Do you have any comment on the things I fixed here?

src/index.js
```diff
@@ -132,7 +126,8 @@ Inspector.prototype = {
       helper.visible = false;
       this.sceneHelpers.add(helper);
       this.helpers[object.uuid] = helper;
-      helper.update();
+      // SkeletonHelper doesn't have an update method
+      if (helper.update) helper.update();
```
dmarcos: add brackets to the if body

vincentfretin: done
I'll fix other ones that eslint reports in the code base as part of #639

vincentfretin:
Another UI issue fixed in #648

vincentfretin:
For the animation loop fix. Before you had a NaN in the panel for the loop property when clicking on the red cube, see:
Capture d’écran du 2022-10-10 15-33-36
With the fix, you have properly a checkbox and you can disable the animation loop.

vincentfretin:
For the "entityclone" fix. It fixes the left panel not refreshing when you click on the Clone entity icon here:
Capture d’écran du 2022-10-10 15-36-48

vincentfretin:
The `event.preventDefault();` fixes are to prevent panel to collapse/expand, this prevents the default behavior that is following the defined href. So before the url address bar would change to add a "#" and wrongly collapse the panel. After the fix, it just copy to clipboard, that's all.

vincentfretin:
I fixed another issue in `lib/viewport.js` with helper update error while trying to expand an entity on the left side panel.
image

dmarcos:
Thank you!

PR #639, opened 2022-10-06
Some code cleanup

vincentfretin:
- Remove dead code, unused imports, unused variables.
- Use import/export everywhere, latest wepback/babel don't allow to use module.exports if you used one import statement.
- Specify a prettier config with option singleQuote to match the current code. I didn't reformatted the files with prettier in this PR.
- Disable the eslint space-before-function-paren rule that conflicts with how prettier format the files.
- Fix eslint errors.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,4 @@
+{
+  "printWidth": 80,
+  "singleQuote": true
+}
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,5 @@
 {
   "printWidth": 80,
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "none"
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@
   "rules": {
     "react/jsx-indent-props": [2, 2],
     "semi": [2, "always"],
+    "space-before-function-paren": "off",
     "no-console": "off"
   },
   "env": {
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+src/lib/TransformControls.js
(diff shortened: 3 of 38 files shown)


vincentfretin:
eslint errors down to 15 errors instead of 350 errors

vincentfretin:
I'll fix the remaining eslint errors later then this PR should be good to merge if there is no objection.

vincentfretin:
I fixed all remaining eslint issues, excepted those:
`npm run lint`
```
src/components/components/Mixins.js
  20:5  error  Do not use setState in componentDidUpdate  react/no-did-update-set-state
```
I'll fix this one as part of the react update in #641.

```
src/index.js
  104:11  error  'geometry' is defined but never used  no-unused-vars
  105:11  error  'material' is defined but never used  no-unused-vars
```
This is already fixed in #638 

This PR is ready to be merged after #638.

vincentfretin:
`npm run prettier` will reformat those files
```
	modifié :         src/components/Main.js
	modifié :         src/components/components/CommonComponents.js
	modifié :         src/components/components/Component.js
	modifié :         src/components/scenegraph/Toolbar.js
	modifié :         src/components/viewport/CameraToolbar.js
	modifié :         src/lib/EditorControls.js
	modifié :         src/lib/cameras.js
	modifié :         src/lib/raycaster.js
	modifié :         src/lib/viewport.js
```
but we can do that in a separate PR, I created the issue #642 for this.

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "printWidth": 80,
```
dmarcos, line 2: Does this enforce 80 character lines? If that's the case not sure if these days makes much sense. In today's displays aspect ratio vertical space is more of a premium than horizontal

vincentfretin: It's actually the default.

vincentfretin: All files are already using the 80 characters limit actually, so no real change here.

dmarcos:
This one is ready to go? Thanks for the patience.

vincentfretin:
Yes, please merge.

PR #640, opened 2022-10-06
Fix style issues with bootstrap or tailwind in the site

vincentfretin:
Specify default css properties on some elements to fix style issues when using the inspector in a site that has tailwind base.css or bootstrap reset styles.

Before:
Capture d’écran du 2022-10-04 20-35-46

After:
Capture d’écran du 2022-10-06 15-34-20

Remaining fixes to do:
- [ ] local checkbox and other checkboxes for boolean properties is always black
- [ ] the id/class on the top bar is not aligned with `<a-entity`
- [ ] in the outliner the line height change slightly if I remove or add class

### DIFF
--- a/src/style/components.styl
+++ b/src/style/components.styl
@@ -49,6 +49,7 @@ div.vec4
 .collapsible .static
   background $bglight
   border-bottom 2px solid $bg
+  box-sizing content-box
   cursor pointer
   height 16px
   padding 8px 10px 12px 10px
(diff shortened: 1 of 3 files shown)


vincentfretin:
tailwind css includes base.css that reset some default browser css, like this one:
```
*, ::before, ::after {
  box-sizing: border-box;
}
```

dmarcos:
Thanks!

vincentfretin:
Thanks for merging. I didn't make this PR a draft PR? I'll create a second PR when I can spend time to look for the remaining issues I noted.

PR #644, opened 2022-10-08
Be able to easily create a build without Google Analytics

vincentfretin:
For GDPR compliance, you may want to create a build without Google Analytics. I added condition around the `ga()` calls to be able to easily create a build without Google Analytics but just commenting the `require('../vendor/ga');` in `index.js`.


vincentfretin:
@dmarcos You can merge this one alone, otherwise the changes are also in #647

dmarcos:
Thank you very much

PR #598, opened 2020-03-07
Fix inspector toggle on US international keyboards

svenkonings:
Fixes issue #595 
See also issue aframevr/aframe#4089

On US international keyboards Ctrl + Alt are combined into AltGraph. When this happens Ctrl and Alt are registered as unpressed, breaking the toggle expression

This is fixed by adding an additional clause for the AltGraph combination.

PR #658, opened 2022-12-06
Fix inspector toggle on US international keyboards

vincentfretin:
This replaces #598 with the correct prettier formatting. It was part of #647


dmarcos:
Thank you!

PR #667, opened 2022-12-24
Fix remaining style issues with tailwindcss in the site

### LINKED ISSUE #663: Remaining style issues with tailwindcss in the site
Remaining todos that was listed in #640
- local checkbox and other checkboxes for boolean properties is always black
- the id/class on the top bar is not aligned with <a-entity
- in the outliner the line height change slightly if I remove or add class

vincentfretin:
This closes #663

Capture d’écran du 2022-12-24 14-28-16

### DIFF
--- a/src/style/components.styl
+++ b/src/style/components.styl
@@ -146,6 +146,10 @@ div.vec4
     box-sizing border-box
     width 165px
 
+  input[type=text]:focus,
+  input.string:focus
+    box-shadow none
+
   .color_value
     margin 0 0 0 5px
     width 68px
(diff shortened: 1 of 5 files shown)


vincentfretin:
Fix initial value position in select fields:
```css
.select__input
  min-height auto !important
```
This is to override the following rule in `components.styl`
```css
  input[type=text],
  input[type=number],
  input.string,
  input.number
    min-height 26px
```

Before:
Capture d’écran du 2022-12-24 15-26-34
After:

Capture d’écran du 2022-12-24 15-25-49

dmarcos:
Thank you!

PR #669, opened 2022-12-30
Cancel tailwindcss default style on select input when adding a component

vincentfretin:
I missed a case when I did #667

Capture d’écran du 2022-12-30 14-03-37
The change remove the blue border in the screenshot.


dmarcos:
Thank you!

PR #670, opened 2022-12-30
Really use the newly active camera when exiting edit mode

vincentfretin:
The code didn't make sense, it didn't do anything. It didn't do what the comment says, now it does.
You can test it:
- go in edit mode
- add a new entity and add a camera component, this camera will be active, the previous one has active unchecked.
- click on "Back to scene" button

The new active camera is correctly used.


dmarcos:
Thank you!

PR #678, opened 2023-02-14
Use Object.is to handle NaN comparison in NumberWidget to avoid infinite React update (fix #677)

### LINKED ISSUE #677: Crashes when physics object selected
https://stackoverflow.com/questions/75306826/a-frame-inspector-disappears/75418222#75418222

Getting React error:
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

klyap:
Problem:
Selecting an object in the scene with ammo physics applied crashes the inspector. 
Inspector also crashes if you happen to pass in something like "fog: false" for `aframe-environment-component` because it expects a number.

Steps to repro:
Open the example/index.html file in this repo. Click on the environment object in the scene, which happens to have the offending "fog: false" param. Inspector will crash.

Cause:
This happens when a component property value comes in as NaN. For example, somehow the `ammo-body` component's schema default value is `NaN`, and when this is passed through into `NumberWidget.js`, results in an infinite update because it thinks NaN !== NaN. The "fog: false" example also sends `NaN` as the value for `fog`, resulting in the same error.

Fix:
I swapped in `isObject` for a more robust check (will give true when comparing 2 NaNs) so that this infinite state update doesn't happen for similar scenarios.

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -102,3 +102,15 @@ export function saveBlob(blob, filename) {
   link.click();
   // URL.revokeObjectURL(url); breaks Firefox...
 }
+
+// Compares 2 vector objects up to size 4
+// Expect v1 and v2 to take format {x: number, y: number, z: number, w:number}
+// Smaller vectors (ie. vec2) should work as well since their z & w vals will be the same (undefined)
+export function areVectorsEqual(v1, v2) {
+  return (
+    Object.is(v1.x, v2.x) &&
+    Object.is(v1.y, v2.y) &&
+    Object.is(v1.z, v2.z) &&
+    Object.is(v1.w, v2.w)
+  );
+}
--- a/src/components/widgets/NumberWidget.js
+++ b/src/components/widgets/NumberWidget.js
@@ -113,6 +113,9 @@ export default class NumberWidget extends React.Component {
   componentDidUpdate(prevProps) {
     // This will be triggered typically when the element is changed directly with
     // element.setAttribute.
+
+    // We use Object.is instead of === for comparison here so that comparing two NaN doesn't trigger an infinite update.
+    // Object.is(NaN, NaN) is true, NaN === NaN is false
     if (!Object.is(this.props.value, prevProps.value)) {
       this.setState({
         value: this.props.value,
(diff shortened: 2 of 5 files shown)


vincentfretin:
Thank you so much @klyap for the investigation and the fix. This fixes #677
Can you please edit your commit message to include the issue number and be more specific about NaN, something like this:
`Use Object.is to handle NaN comparison in NumberWidget to avoid infinite React update (fix #677)`.
You can also add a comment in the code, something like:
 `// We use Object.is instead of === for comparison here so that comparing two NaN doesn't trigger an infinite update. Object.is(NaN, NaN) is true, NaN === NaN is false`

klyap:
@vincentfretin made the changes! thanks for reviewing!

src/lib/utils.js
```diff
@@ -102,3 +102,15 @@ export function saveBlob(blob, filename) {
   link.click();
   // URL.revokeObjectURL(url); breaks Firefox...
 }
+
+// Compares 2 vector objects up to size 4
+// Expect v1 and v2 to take format {x: number, y: number, z: number, w:number}
+// Smaller vectors (ie. vec2) should work as well since their z & w vals will be the same (undefined)
+export function areVectorsEqual(v1, v2) {
```
klyap, line 109: Made a util function for this since it's used in all 3 Vec Widget files.

vincentfretin:
Thanks @klyap. Looks good to me.
@dmarcos you can merge this if the changes are ok for you.

src/lib/utils.js
```diff
@@ -102,3 +102,15 @@ export function saveBlob(blob, filename) {
...
+  return (
+    Object.is(v1.x, v2.x) &&
```
dmarcos, line 111: I would probably use the usual `v1.x === v2.x`

dmarcos:
Thanks so much. Just a nit and we can merge

src/lib/utils.js
vincentfretin: No, that's the point of the fix here.
We use Object.is instead of === for comparison here so that comparing two NaN doesn't trigger an infinite update. Object.is(NaN, NaN) is true, NaN === NaN is false

vincentfretin: We added the above comment in NumberWidget where we also used Object.is. See the context here
https://github.com/aframevr/aframe-inspector/issues/677#issuecomment-1429366813

dmarcos:
Thanks!

PR #680, opened 2023-03-14
Fix panels disappearing when selecting a-videosphere entity

vincentfretin:
See details of the debugging issue in https://github.com/aframevr/aframe/issues/5257
This fixes panels disappearing when selecting a-videosphere entity
Capture d’écran du 2023-03-14 11-31-44

### DIFF
--- a/examples/360video.html
+++ b/examples/360video.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>360 Video</title>
+    <meta name="description" content="360 Video — A-Frame">
+    <script src="https://cdn.jsdelivr.net/gh/aframevr/aframe@master/dist/aframe-master.min.js"></script>
+    <!--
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks so much! Do we need to commit the dist?

vincentfretin:
It's good to have an updated dist on master yes, this way we can use 
```<a-scene inspector="url: https://cdn.jsdelivr.net/gh/aframevr/aframe-inspector@master/dist/aframe-inspector.min.js">```
without waiting for a 1.4.2 release.

vincentfretin:
But your call really. I can remove the commit if you want. And you do the dist and commit it yourself, right after the merge or wait for a new release. Let met know.

dmarcos:
Problem is that it makes the PRs noisy. Maybe remove for now and can think of setting a build bot like in A-Frame

vincentfretin:
I removed the dist.

dmarcos:
Thanks so much!

PR #682, opened 2023-03-31
Update link to aframe-watcher

vincentfretin:
Replace the supermedium.com/aframe-watcher link by https://github.com/supermedium/aframe-watcher


dmarcos:
Thank you very much

Issue #649, opened 2022-10-17
Transform controls really big


vincentfretin:
There is an issue where the transform controls doesn't resize when zooming in or out, so it appears really big when the camera is far away and you focus on a small object (with f or dblclick).
![transform_controls_really_big](https://user-images.githubusercontent.com/112249/196130306-1223e224-663e-43cf-b77b-bd17ddc9a3a3.jpg)

vincentfretin:
Normal size if I click away and reclick on the object:

![transform_controls_normal_size](https://user-images.githubusercontent.com/112249/196130788-9319252d-d5f4-4379-b767-3d4441cd70fe.jpg)

dmarcos:
Any quick fix for this we can incorporate to 1.4.0?

vincentfretin:
This is most probably an update method to call after using zoom.
I quickly look at it but I'm not familiar enough with the code to do a quick fix on this.
I plan to work on updating TransformControls next year, we should be able to use the latest one from three instead of our very old one in the repository (for example setScaleSnap is missing on our old version).

Also I see that in EditorControls.js we're patching `THREE.Box3.prototype.expandByObject` so overriding the one from three, that's not great.

Also I would disable the dbclick to focus behavior, it's really confusing when you do it by mistake on the ground plane, it zooms out to infinity.

vincentfretin:
You can sponsor [vincentfretin](https://github.com/sponsors/vincentfretin) to expedite this fix. Find other work you can sponsor at https://github.com/c-frame/sponsorship

PR #686, opened 2023-04-17
Remove Box3.prototype.expandByObject patch

vincentfretin:
Remove `Box3.prototype.expandByObject` patch, this function is an old version of what now exists in latest threejs:
https://github.com/mrdoob/three.js/blob/25859b6bdf64a13301dffdad6369b0c675584aa4/src/math/Box3.js#L155-L220

I mentioned it in my comment https://github.com/aframevr/aframe-inspector/issues/649#issuecomment-1364654680
It's been months since I commented that code in my project to not patch Box3 with an old version of expandByObject. I didn't see any issue.

I verified for example with the blue cube with a sphere as a child
```
<a-entity id="blueBox" mixin="blueBox" position="0 8 0">
  <a-entity id="yellowSphere" geometry="primitive: sphere" material="color:#ff0; metalness:0.0; roughness:1.0" position="-2 2 -2"></a-entity>
</a-entity>
```
This works properly:
Capture d’écran du 2023-04-17 15-47-53

### DIFF
--- a/src/components/widgets/NumberWidget.js
+++ b/src/components/widgets/NumberWidget.js
@@ -92,6 +92,11 @@ export default class NumberWidget extends React.Component {
         value = parseFloat(value);
       }
 
+      // If we inadvertently typed a character in the field, set value to the previous value from props
+      if (isNaN(value)) {
+        value = this.props.value;
+      }
+
       if (value < this.props.min) {
         value = this.props.min;
       }
--- a/src/lib/EditorControls.js
+++ b/src/lib/EditorControls.js
@@ -1,57 +1,5 @@
 import debounce from 'lodash.debounce';
 
-THREE.Box3.prototype.expandByObject = (function () {
-  // Computes the world-axis-aligned bounding box of an object (including its children),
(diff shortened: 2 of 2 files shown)


vincentfretin:
I saw recently how I can have a NaN for position.
There a small difference between having the patch or not.
If you type inadvertently some characters for position that are not a digit, you will get a NaN in the field.
With the patch, no error in the console and you have the previous blue bounding box still there but not the model.
Without the patch, the model and blue bounding box disappear and you can the following error in the console:
`THREE.BufferGeometry.computeBoundingSphere(): Computed radius is NaN. The "position" attribute is likely to have NaN values.`

I still believe the patch should be removed.
Another fix could be done to avoid the field to have NaN and put back the previous value or 0.

vincentfretin:
We definitely need to remove this old expandByObject patch, this fixed several issues that 3DStreet was having with some of their models https://github.com/3DStreet/3dstreet-editor/pull/383

I rebased from master and fixed the issue of having a NaN in a number input like position if you typed a character inadvertently, putting the previous value given in props.

Issue #676, opened 2023-01-25
suggestion: remove texture dialog


kfarr:
Texture modal dialog is the cause of many issues on github:
https://github.com/aframevr/aframe-inspector/issues/429
https://github.com/aframevr/aframe-inspector/issues/471
https://github.com/aframevr/aframe-inspector/issues/494
https://github.com/aframevr/aframe-inspector/issues/508
https://github.com/aframevr/aframe-inspector/issues/621

it also appears to re-download textures when i look at network activity, effectively doubling download bytes.

Would it make sense to remove this feature?

vincentfretin:
If the direction that @dmarcos want this repo to take is to stay an inspector and not add any new features to be more an editor, it would make sense to drop this feature indeed, and also remove the aframe-watcher button #679 
We can implement a better editor in a fork, I started one in https://github.com/c-frame/aframe-editor to add drag an drop, but no specific roadmap for now.

vincentfretin:
The only open issues remaining are:
- #429 
- #599

otherwise the texture modal is working properly.

When you select a new texture, it indeed does the query twice, once to create the asset in a-assets and another one to create an image to render in the modal, but with "Disable cache" unchecked, it says (disk cache) so that's not really an issue as far as I can tell.

PR #691, opened 2023-07-09
Delete aframe.js (ancient & unused copy of code from 0.4.0)

diarmidmackenzie:
Following on from https://github.com/aframevr/aframe/pull/5333 and based on comments from @vincentfretin, I checked aframe-inspector for any usage of 'https://cdn.aframe.io'.

The only instances I found were in the module examples\js\aframe.js.

This appears to be a copy of the 0.4.0 distribution of A-Frame, and not used by any of the examples.  I believe it can be deleted, and it would be good housekeeping to do so.

Hence this PR.


vincentfretin (approved):
Yes indeed.

vincentfretin:
The list of textures images I talked about in comment https://github.com/aframevr/aframe/issues/5119#issuecomment-1382752776 is retrieved via https://aframe.io/sample-assets/dist/images.json when you open the texture modal.
The url is defined here https://github.com/aframevr/aframe-inspector/blob/72fd46159801ad37a14e87448611607b5d09a36e/src/lib/assetsLoader.js#L3 but that's different from https://cdn.aframe.io/ so you can't use AFRAME_ROOT_CDN here. If we introduce another configurable constant, that can be done in another PR. But also this feature should maybe be dropped completely, see #676

vincentfretin:
For my comment above, I created #702 

@dmarcos this PR just remove an unused file, you can merge.

dmarcos:
Thanks!

PR #695, opened 2023-07-22
Fix duplicate entity feature not using the modified value for position/rotation/scale

### LINKED ISSUE #688: flushtodom() called during inspector session outputs old values (values prior to inspector changes)
This is the second of 2 errors encountered using the duplicate (clone) entity feature (keyboard shortcut `d`)

* when an entity is cloned, first the entity's current values are flushed to the DOM since the clone feature uses the vanilla javascript method `cloneNode` to clone an entity.
* when flushToDOM is executed, the values returned are those of the entity prior to invoking the inspector. In other words, if I change position or other attribute using the editor, upon pressing `d` to duplicate, the entity will return to its prior location before making the changes in the editor.

here is the flushtodom for an entity:
https://github.com/aframevr/aframe/blob/v1.4.0/src/core/a-entity.js#L698

for each component in the entity, it runs this flushtodom:
https://github.com/aframevr/aframe/blob/v1.4.0/src/core/component.js#L267

note how it is using `this.attrValue` instead of `getAttribute([componentName], [attributeName])` which may be returning the values in the original html instead of the current values using A-Frame's recommended api

here is a video
https://github.com/aframevr/aframe-inspector/assets/470477/b7cb3adb-1408-4709-9cd2-3a65b4990ca4

here is the URL to reproduce:
https://mixin-gltf-url-error.glitch.me/

here are instructions to reproduce:
press ctl-alt-i to invoke the inspector, change the position of the mixin firetruck, while the truck entity is still selected press `d`, note how the firetruck returns to its original position

expected behavior:
- when cloning an item the original and the duplicated item retain attribute values modified during the current inspector session

vincentfretin:
Properly call `setAttribute` for position, rotation, scale when modifying with the TransformControls, so that `entity.flushToDOM()` works correctly when duplicating an entity.
This fixes #688.
See my comment https://github.com/aframevr/aframe-inspector/issues/688#issuecomment-1646632546 for a more detailed explanation of the issue and the proposed fix.

### DIFF
--- a/src/lib/viewport.js
+++ b/src/lib/viewport.js
@@ -74,6 +74,11 @@ export function Viewport(inspector) {
       component = 'scale';
       value = `${object.scale.x} ${object.scale.y} ${object.scale.z}`;
     }
+
+    // We need to call setAttribute for component attrValue to be up to date,
+    // so that entity.flushToDOM() works correctly when duplicating an entity.
+    transformControls.object.el.setAttribute(component, value);
+
     Events.emit('entityupdate', {
       component: component,
       entity: transformControls.object.el,


dmarcos:
Thanks!

vincentfretin:
@dmarcos You can also close https://github.com/aframevr/aframe/issues/4289 and https://github.com/aframevr/aframe/issues/4084 that were the same issue.

PR #700, opened 2023-12-07
Remove Google Analytics

vincentfretin:
Remove Google Analytics, that's legacy @dmarcos said.
Also a step further to GDPR compliance.


dmarcos:
Thank you!

PR #706, opened 2024-02-04
Replace fontawesome cdn by react-fontawesome

vincentfretin:
This closes #692
Some icons slightly changed.

Remaining to do:
- [x] Regression for the collapsible panel when clicking on clipboard icon
- [x] save icon
- [x] icons for types camera: 'fa-camera', mesh: 'fa-cube', light: 'fa-lightbulb-o', text: 'fa-font'

Issue #692, opened 2023-07-10
Use icons from @fortawesome/react-fontawesome


vincentfretin:
For GDPR reason to have one less external url to fetch resources from, and as a bonus reduce size of downloaded assets, replace usage of https://maxcdn.bootstrapcdn.com used for fontawesome icons 
https://github.com/aframevr/aframe-inspector/blob/72fd46159801ad37a14e87448611607b5d09a36e/src/components/Main.js#L14-L17

by [@fortawesome/react-fontawesome](https://www.npmjs.com/package/@fortawesome/react-fontawesome) npm package.
That's it, instead of using class "fa fa-pause" on an element, you use
```js
import { faPause } from "@fortawesome/free-solid-svg-icons";
import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
<FontAwesomeIcon icon={faPause} />
```

PR #707, opened 2024-02-26
Reduce bundle size after the fontawesome icons changes

vincentfretin:
This addresses my comment https://github.com/aframevr/aframe-inspector/pull/706#issuecomment-1952042203 about he unfortunate increase of the bundle size.
When I wanted to do #692 I had in mind a decrease of what need to be downloaded, not an increase. This PR fixes it.

Removing usage of FontAwesomeIcon component and using our own simplified component remove 66 kB.
I also switched from classnames to lighter clsx that is used nowadays in React projects, removing 1 kB.
This PR reduces the minified bundle from 554K to 489K.

Before #706 we had 483K+27K=510K
After #706 and this PR: 489K.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,8 @@
       "license": "MIT",
       "dependencies": {
         "@fortawesome/free-solid-svg-icons": "^6.5.1",
-        "classnames": "^2.3.2",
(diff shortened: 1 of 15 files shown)


dmarcos:
Thanks so much!

PR #708, opened 2024-02-28
Update build dependencies, three and react-select

vincentfretin:
Mainly maintenance update just to keep things up to date, while I'm doing this in all my other projects.
Please not that some build dependencies update require now node 18.x (previous nodes version are not maintained anyway, node 20.x is the current LTS)
Updating three from 0.145.0 to 0.161.0 for GLTFExporter, exporting the example scene still works great, no change, bundle size increased by 7kB because of this update.
Updating react-select from 5.7.0 to 5.8.0, no real change for the usage we do here.

Bundle size
before: 489K
after: 493K


vincentfretin:
@dmarcos there is still a very old and unmaintained ghpages package in the dependencies. I don't think we use it. Do you use any of the npm run ghpages/preghpages/deploy scripts when you update https://aframe.io/aframe-inspector/examples/ ?
If not, we can remove that stuff.

dmarcos:
thanks!

PR #711, opened 2024-05-22
Wait for gltf-model to be loaded to show the selectionBox

vincentfretin:
When you create an entity with the entitycreate event with components object having a gltf-model, the selectionBox didn't appear. This fixes it.


dmarcos:
Thanks

PR #712, opened 2024-05-22
Add optional support for id and class in createEntity definition

vincentfretin:
Add optional support for id and class in createEntity definition make element optional, default to a-entity.


dmarcos:
Thank you! You rock!

PR #714, opened 2024-05-22
Support for translationsnapchanged and rotationsnapchanged

vincentfretin:
Rename snapchanged to translationsnapchanged and add rotationsnapchanged to be able to configure snapping for both translation and rotation.
The current React interface doesn't use this code.


dmarcos:
Thanks!

PR #717, opened 2024-05-23
Keep the camera position when switching to inspector (fix #607)

### LINKED ISSUE #607: Be able to look through the cameras in the scene
I want to go into the Visual Inspector, but I want to keep the same perspective I had before I opened the inspector (i.e. looking through the camera in the scene). As far as I can tell, the default behavior is to relocate the viewport to a “neutral” location called "perspective". 

What I’d love would be if the view menu included all of the cameras in the scene as well as the generic “perspective”; this would be a nice general solution.

vincentfretin:
Keep the camera position and rotation when opening the inspector.

I have this change for quite some times in my project. When I use the default inspector build in other projects I am so frustrated that the camera position is not kept while opening the inspector.
It seems I am not the only one that wanted this behavior.

This closes #607

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,3 +68,15 @@ edit `index.html` and define before any script tag this global variable:
 ```html
 <script>window.AFRAME_SAMPLE_ASSETS_ROOT = "./sample-assets/";</script>
 ```
+
+## Config overrides
+
+Since A-Frame 1.7.0, the inspector perspective camera position is kept in sync with the A-Frame
+active camera. This means you can move around the scene, toggle the inspector and you will be at the same position.
+If you want to disable that behavior, you can do that by defining a global variable like this:
+
+```html
+<script>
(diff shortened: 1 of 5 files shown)


vincentfretin:
But it should be an option somehow. I want that behavior when I am on a scene and I open the inspector.
But I don't want that behavior when opening directly the inspector and want the perspective camera to be at the specified position https://github.com/aframevr/aframe-inspector/blob/575c9106e26b41c714cc369b55c4db861315d7a0/src/lib/cameras.js#L45-L46
Those positions should also be configurable, with the ortho cameras as well. I'm writing some configurators where I want different camera positions.

vincentfretin:
I made the new behavior of keeping the inspector perspective camera position in sync with the A-Frame active camera the default, with the ability to disable that behavior with a window.AFRAME_INSPECTOR_CONFIG variable.

PR #718, opened 2024-06-04
Add functions to emit and listen events of inspector

gregogalante:
Sometimes we need to manage inspector functionalities from the outside. We can access with AFRAME.INSPECTOR but some operations can not be done with the Inspector class instance.

With this mode we can listen and emit events using AFRAME.INSPECTOR.


vincentfretin:
There is already https://github.com/aframevr/aframe-inspector/blob/575c9106e26b41c714cc369b55c4db861315d7a0/src/index.js#L20 exposed
We can just add `this.off = Events.off;` instead of using two other functions.
Please don't include the dist in the PR.

vincentfretin:
Ah and `this.emit = Events.emit` of course.

gregogalante:
@vincentfretin this.emit = Events.emit is not existing. I create a new pull request

vincentfretin:
Yeah this is what I meant, change your PR to expose `off` and `emit`.

vincentfretin:
This was not working. It will be reverted in #755

PR #719, opened 2024-06-04
Added this.emit = Events.emit to add possibility to emit custom events from outside the Inspector class.

gregogalante:
Added this.emit = Events.emit to add possibility to emit custom events from outside the Inspector class.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ function Inspector() {
   this.isFirstOpen = true;
   this.modules = {};
   this.on = Events.on;
+  this.emit = Events.emit;
   this.opened = false;
 
   // Wait for stuff.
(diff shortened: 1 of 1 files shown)


vincentfretin:
It would make sense to also expose `this.off = Events.off;` so you can remove a listener you added previously with AFRAME.INSPECTOR.on. (Ok just saw you added it)

vincentfretin:
I would be interested of knowing your use case with this, creating entity with the entitycreate event maybe? You can DM me on discord (handle vincentfretin) for a quick chat.
I myself have the whole aframe-inspector code in my own repo under src/editor. This is also what 3dstreet is doing now, so we're just importing Events. We're also working on a new json import/export of the scene, see https://github.com/3DStreet/3dstreet/pull/537

dmarcos:
Thanks so much. Congrats on your first contribution @gregogalante

vincentfretin:
Did you forgot to push the Merge button @dmarcos?

dmarcos:
thanks

vincentfretin:
This was not working. It will be reverted in #755

PR #720, opened 2024-06-05
Fix issue with zoom with mouse wheel jumping when switching camera

vincentfretin:
Fix issue with zoom with mouse wheel jumping when switching to an ortho camera and zooming because of missing `camera.updateProjectionMatrix()`


dmarcos:
Thanks!

PR #724, opened 2024-06-16
Remove extra space when we use search in left panel

vincentfretin:
That was a minor regression when I did #706

image

Before
image

After
image


vincentfretin:
@dmarcos in case you missed the notification. There are also other PRs.

dmarcos:
Thanks

PR #725, opened 2024-06-16
Fix blue hover on all icon buttons

vincentfretin:
Remove blue on hover on the icon in entity representation in right panel
Add blue hover on toolbarActions icons (plus, play, save)
Fix button css rules for modal to properly style the close button


vincentfretin:
The changes of the css rules for the modal is just fixing the rules to add the missing ":". I guess those were lost with the switching to stylus years ago. @dmarcos

dmarcos:
Thank you!

PR #728, opened 2024-06-28
Don't trigger onChange when mounting NumberWidget component

vincentfretin:
When clicking an entity that have a position with 5 decimals, it opens the right panel, triggering props.onChange on the NumberWidget react component for position x y z, thus truncating to 3 decimals the coordinates and adding a change to the undo stack I'm currently implementing. We don't want to do any change when just mounting the react component.

### DIFF
--- a/src/components/widgets/NumberWidget.js
+++ b/src/components/widgets/NumberWidget.js
@@ -38,9 +38,6 @@ export default class NumberWidget extends React.Component {
     this.distance = 0;
     this.onMouseDownValue = 0;
     this.prevPointer = [0, 0];
-
-    this.setValue(this.props.value);
-    this.onBlur();
   }
 
   onMouseMove = (event) => {
@@ -131,7 +128,6 @@ export default class NumberWidget extends React.Component {
 
   onBlur = () => {
     this.setValue(parseFloat(this.input.current.value));
-    this.setState({ class: '' });
   };
 
   onChange = (e) => {
@@ -143,7 +139,6 @@ export default class NumberWidget extends React.Component {
 
     // enter.
     if (event.keyCode === 13) {
-      this.setValue(parseFloat(this.input.current.value));
       this.input.current.blur();
       return;
     }
@@ -171,7 +166,6 @@ export default class NumberWidget extends React.Component {
         onKeyDown={this.onKeyDown}
         onChange={this.onChange}
         onMouseDown={this.onMouseDown}
-        onFocus={this.onFocus}
         onBlur={this.onBlur}
       />
     );


src/components/widgets/NumberWidget.js
```diff
@@ -38,9 +38,6 @@ export default class NumberWidget extends React.Component {
     this.distance = 0;
     this.onMouseDownValue = 0;
     this.prevPointer = [0, 0];
-
-    this.setValue(this.props.value);
```
vincentfretin, line 42: setValue here was a noop, because we have `if (value === this.state.value) return;` in setValue, and this.state.value is already this.props.value from constructor.

src/components/widgets/NumberWidget.js
```diff
@@ -131,7 +128,6 @@ export default class NumberWidget extends React.Component {
 
   onBlur = () => {
     this.setValue(parseFloat(this.input.current.value));
-    this.setState({ class: '' });
```
vincentfretin, line 134: dead code, this.state.class is not used in this component

src/components/widgets/NumberWidget.js
```diff
@@ -38,9 +38,6 @@ export default class NumberWidget extends React.Component {
     this.distance = 0;
     this.onMouseDownValue = 0;
     this.prevPointer = [0, 0];
-
-    this.setValue(this.props.value);
-    this.onBlur();
```
vincentfretin, line 43: This is the issue, truncating 5 decimals to 3 decimals and triggering the props.onChange callback.

src/components/widgets/NumberWidget.js
```diff
@@ -143,7 +139,6 @@ export default class NumberWidget extends React.Component {
 
     // enter.
     if (event.keyCode === 13) {
-      this.setValue(parseFloat(this.input.current.value));
```
vincentfretin, line 146: This is already done by the onBlur callback that's called with the blur() call below.

src/components/widgets/NumberWidget.js
```diff
@@ -171,7 +166,6 @@ export default class NumberWidget extends React.Component {
         onKeyDown={this.onKeyDown}
         onChange={this.onChange}
         onMouseDown={this.onMouseDown}
-        onFocus={this.onFocus}
```
vincentfretin, line 174: dead code, this.onFocus is undefined

vincentfretin:
Note that when you change the value with the input field, it keeps the 5 decimals precision while using arrows but it truncates to 3 decimals on blur at the end. Only while using mouse that the precision is kept at 5 decimals because the input doesn't have the focus, so not blur.
It would probably make sense to use 3 decimals instead of 5 for TransformControls as well.

vincentfretin:
I need this PR to be merged before doing the undo position/rotation/scale PR @dmarcos

dmarcos:
Thanks!

PR #729, opened 2024-06-28
Make ctrl+x/c/v/s work on linux and macOS

vincentfretin:
We also use ctrl key on Ubuntu, not just Windows ;)

Those are the shortcuts to cut, copy, paste an entity and focus the search input.

### DIFF
--- a/src/lib/shortcuts.js
+++ b/src/lib/shortcuts.js
@@ -130,7 +130,7 @@ export const Shortcuts = {
     }
 
     if (
-      (event.ctrlKey && os === 'windows') ||
+      (event.ctrlKey && os !== 'macos') ||
       (event.metaKey && os === 'macos')
     ) {
       if (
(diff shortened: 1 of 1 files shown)


vincentfretin:
I need this PR to be merged before doing the undo position/rotation/scale PR @dmarcos

vincentfretin:
Actually it didn't work either on macOS because of metaKey check in shouldCaptureKeyEvent.
So it only worked on Windows previously.

I checked with @kfarr for macOS for all the shortcuts here, we don't see why there was the metaKey check in shouldCaptureKeyEvent here. We just know the implementation was initially copied from AFRAME.utils.shouldCaptureKeyEvent that is used mainly for wasd-controls, but even for this I'm not sure why we need to check for metaKey (cmd key on macOS).

dmarcos:
Thanks!

PR #730, opened 2024-06-28
[cleanup] Properly remove prop for multi-prop component when value is null or undefined

vincentfretin:
In updateEntity function used on the onChange callback for all widgets,
properly remove prop for multi-prop component when value is null or undefined, the previous code didn't do anything.
Currently none of the widgets onChange seems to go in that branch but this should still be fixed.


dmarcos:
Thanks!

PR #731, opened 2024-06-29
[feature] Add help button on the right of toolbar actions to show shortcuts modal

vincentfretin:
The h shortcut to show the shortcuts modal is not known by many. Although this is specified in the documentation here
https://aframe.io/docs/master/introduction/visual-inspector-and-dev-tools.html#shortcuts not a lot of persons read the full documentation, or if one reads it, they probably won't remember that this shortcut exists. I actually discovered the feature several years later of developing with aframe when I started updating aframe inspector code.
To make it more discoverable, I propose to add a help button on the right of toolbar actions to show the shortcuts modal.
See the screenshot:
image

### DIFF
--- a/src/components/scenegraph/Toolbar.js
+++ b/src/components/scenegraph/Toolbar.js
@@ -3,7 +3,8 @@ import {
   faPlus,
   faPause,
   faPlay,
-  faFloppyDisk
+  faFloppyDisk,
+  faQuestion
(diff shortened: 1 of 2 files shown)


vincentfretin:
What do you think @dmarcos?

dmarcos:
Thank you!

PR #733, opened 2024-07-15
[bugfix] Update selection box on entityupdate only if it's the selected entity

vincentfretin:
This fixes an error when using undo with unselected entity so when inspector.selectedEntity can be undefined.


vincentfretin:
This is a required change for undo feature @dmarcos 
The condition is similar to the other entityupdate listener we have here
https://github.com/aframevr/aframe-inspector/blob/b792e9113c416f758caccf154acf558c6f538e36/src/lib/viewport.js#L182-L184

dmarcos:
Thanks so much!

PR #735, opened 2024-07-16
[feature] Add a camerachanged event you can use with pathtracer

vincentfretin:
Remove EditorControls debounce on `change` event and add a `camerachanged` event that you can use with [pathtracer](https://github.com/gkjohnson/three-gpu-pathtracer)
This is similar to three.js editor https://github.com/mrdoob/three.js/blob/8a102ea21d19adfd8214e20a347d80f49e098d8c/editor/js/Viewport.js#L280

### DIFF
--- a/src/lib/EditorControls.js
+++ b/src/lib/EditorControls.js
@@ -1,5 +1,3 @@
-import debounce from 'lodash.debounce';
-
 /**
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks!

dmarcos:
Thanks!

PR #736, opened 2024-07-16
[bugfix] Dispose helper when removing it

vincentfretin:
Dispose helper when removing it to properly dispose material and geometry.
In recent three.js version all helpers have now a dispose function, that was not the case historically.
This is similar to three.js editor
https://github.com/mrdoob/three.js/blob/267e76c2d884cb0b4c12f4c5ffa5a7d03ce82bba/editor/js/Editor.js#L472

Also remove the second argument to CameraHelper, CameraHelper only takes one argument.
https://github.com/mrdoob/three.js/blob/267e76c2d884cb0b4c12f4c5ffa5a7d03ce82bba/src/helpers/CameraHelper.js#L21

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -105,7 +105,7 @@ Inspector.prototype = {
       let helper;
 
       if (object instanceof THREE.Camera) {
-        this.cameraHelper = helper = new THREE.CameraHelper(object, 0.1);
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

PR #737, opened 2024-07-16
[cleanup] Remove unused onDoubleClickPosition

vincentfretin:
In three.js editor, this is used https://github.com/mrdoob/three.js/blob/267e76c2d884cb0b4c12f4c5ffa5a7d03ce82bba/editor/js/Viewport.js#L255-L258
but here we don't use it, we use aframe cursor component instead.


dmarcos:
Thanks!

PR #738, opened 2024-07-17
[bugfix] Update camera aspect ratio on window resize

vincentfretin:
Updating the aspect ratio of the editor camera worked only for the active perspective camera, it didn't work when when you were using orthographic camera, and also the case using ortho, resize and then switching back to perspective camera didn't work either. This PR fixes updating the camera aspect in all cases.

For more details on the issue, this was discussed in https://github.com/3DStreet/3dstreet/issues/635

### DIFF
--- a/src/lib/cameras.js
+++ b/src/lib/cameras.js
@@ -118,5 +118,4 @@ function setOrthoCamera(camera, dir, ratio) {
   camera.bottom = info.bottom || -10;
   camera.position.copy(info.position);
   camera.rotation.copy(info.rotation);
-  camera.updateProjectionMatrix();
 }
--- a/src/lib/viewport.js
+++ b/src/lib/viewport.js
@@ -111,6 +111,7 @@ export function Viewport(inspector) {
   Events.on('cameratoggle', (data) => {
     controls.setCamera(data.camera);
     transformControls.setCamera(data.camera);
+    updateAspectRatio();
   });
 
   function disableControls() {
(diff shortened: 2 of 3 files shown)


src/lib/cameras.js
```diff
@@ -118,5 +118,4 @@ function setOrthoCamera(camera, dir, ratio) {
   camera.bottom = info.bottom || -10;
   camera.position.copy(info.position);
   camera.rotation.copy(info.rotation);
-  camera.updateProjectionMatrix();
```
vincentfretin, line 121: This is now done in the updateAspectRatio() call that is done on cameratoggle event.

src/lib/viewport.js
```diff
@@ -202,11 +203,32 @@ export function Viewport(inspector) {
...
+
+    controls.setAspectRatio(aspect); // for zoom in/out to work correctly for orthographic camera
     camera.updateProjectionMatrix();
-  });
+
+    const cameraHelper = inspector.helpers[camera.uuid];
+    if (cameraHelper) cameraHelper.update();
+  }
+
+  inspector.sceneEl.addEventListener('rendererresize', updateAspectRatio);
```
vincentfretin, line 231: I'm using `rendererresize` that is emitted by the `sceneEl.resize()` function. `sceneEl.resize()` is called after switching the active camera when you close or open the inspector.

dmarcos:
Thank you!

PR #740, opened 2024-08-06
[bugfix] Only update selectionBox if the model-loaded event came from the entity

vincentfretin:
Only update selectionBox if the model-loaded event came from the entity and not from a child.
This is a more robust version of what I implemented in #711

I had an issue with a modified version of createEntity that also support creating children with gltf-model.
Before:
Capture d’écran du 2024-08-06 19-19-42
After:
Capture d’écran du 2024-08-06 19-20-21


dmarcos:
Thanks so much

PR #742, opened 2024-08-12
[bugfix] Fix fog NaN in inspector, it should be a number not a boolean

vincentfretin:
The fog property is a number, not a boolean
https://github.com/supermedium/aframe-environment-component/blob/b339b293139cd8c10f03d4c9a162c3b5ea55a794/index.js#L20
I fixed the example.


dmarcos:
Thanks!

PR #744, opened 2024-08-12
[bugfix] Fix React UI crash selecting a-scene in scene graph

### LINKED ISSUE #741: Selecting a-scene crashes the React UI on aframe 1.6.0 and master
image

It works on aframe 1.5.0

vincentfretin:
Check for `undefined` via the optional chaining `?.` operator before accessing the property now that `getDOMAttribute` may return `undefined` in aframe 1.6.0 instead of `{}` we always had previously.

This fixes #741


dmarcos:
Thank you

PR #745, opened 2024-08-16
[cleanup] Remove dead code in several React components

vincentfretin:
Those are props and state that aren't used anywhere.

### DIFF
--- a/src/components/scenegraph/SceneGraph.js
+++ b/src/components/scenegraph/SceneGraph.js
@@ -11,17 +11,13 @@ import Events from '../../lib/Events';
 
 export default class SceneGraph extends React.Component {
   static propTypes = {
-    id: PropTypes.string,
-    onChange: PropTypes.func,
     scene: PropTypes.object,
     selectedEntity: PropTypes.object,
     visible: PropTypes.bool
   };
 
   static defaultProps = {
-    selectedEntity: '',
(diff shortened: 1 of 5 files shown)


src/components/scenegraph/SceneGraph.js
```diff
@@ -70,19 +66,16 @@ export default class SceneGraph extends React.Component {
...
       const entityOption = this.state.filteredEntities[i];
       if (entityOption.entity === entity) {
-        this.setState({ selectedEntity: entity, selectedIndex: i });
+        this.setState({ selectedIndex: i });
         // Make sure selected value is visible in scenegraph
         this.expandToRoot(entity);
-        if (this.props.onChange) {
-          this.props.onChange(entity);
-        }
-        Events.emit('entityselect', entity, true);
```
vincentfretin, line 84: This true param here has no meaning, there is no second parameter in any of the entityselect listener in the code base, meaning there is no `Events.on('entityselect', (entity, someboolean) => ...`

dmarcos:
Thanks!

vincentfretin:
I rebased to fix the conflict. You can merge.

vincentfretin:
@dmarcos in case you missed my comment.

dmarcos:
Thanks!

PR #746, opened 2024-08-16
[bugfix] Check also for curr not null in the while loop

vincentfretin:
Use the ?. operator here to check for undefined or null before accessing isEntity. We got an error at this place one time on 3dstreet because parentNode was null. Note that it makes more sense to check for null here because parentNode can returns null, never undefined normally.


dmarcos:
Awesome

PR #748, opened 2024-08-18
[bugfix] Properly unregister listeners when React component unmount

vincentfretin:
If you toggle the panels with 0 shortcut, the right panel will unmount CommonComponents for example, leaking the entityupdate listener, and registering it again when the React component mount again.
Only CommonComponents and Component were really impacted, the other components doesn't unmount in the current code base, but are generally hidden with css.
I fixed all the React components to properly unregister the listeners on unmount.


dmarcos:
Thank you

PR #749, opened 2024-08-27
[feature] Scroll to selected entity

vincentfretin:
When selecting an entity in viewport or creating new entity with n shortcut, scroll to the selected entity in scene graph.
[scroll-to-selected-entity.webm](https://github.com/user-attachments/assets/253d12ed-6c6f-489a-80bd-785845c07e09)

I'm not sure that's a good behavior for everyone.
Blender for example doesn't have that behavior.
vscode has that behavior when switching tabs, but it doesn't scroll unnecessarily if the other file is already visible in the files explorer. If it isn't visible, it scrolls and the file is in the middle of the files explorer. That may be better.

### DIFF
--- a/src/components/scenegraph/Entity.js
+++ b/src/components/scenegraph/Entity.js
@@ -17,6 +17,7 @@ import Events from '../../lib/Events';
 
 export default class Entity extends React.Component {
   static propTypes = {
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks

PR #750, opened 2024-08-27
[cleanup] Remove unused m shortcut that was used in the removed motion capture feature

vincentfretin:
This should have been removed when the feature was removed in https://github.com/aframevr/aframe-inspector/commit/40730d146b873b205b85957bfd02179a12a2b8da


dmarcos:
Thanks

PR #751, opened 2024-08-27
[cleanup] Simplify updateEntity function

vincentfretin:
Simplify the updateEntity function call for better understanding and accessory avoid creation of string and array.
This will also simplify using the same function for an undo feature.


dmarcos:
Thank you

PR #752, opened 2024-08-27
[cleanup] Remove redundant removeObject call, this is already called on child-detached

vincentfretin:
Remove redundant removeObject call that removes helpers, this is already called on child-detached event.
This was called a second time when removeChild was called because of the listener here
https://github.com/aframevr/aframe-inspector/blob/05d895e62f9319b1479491026888647e4d146f21/src/index.js#L201-L204


dmarcos:
Thank you!

PR #754, opened 2024-08-28
[bugfix] Properly bind the exposed on/off/emit functions

vincentfretin:
Be sure the exposed `on`/`off`/`emit` on `Inspector` are bound to `Events` so an event emitted with `AFRAME.INSPECTOR.emit` is correctly catched in an `Events.on` listener, otherwise it's using two different registries and doesn't work.

And example of usage of emit externally for a component to trigger the update of the scene graph.
https://github.com/3DStreet/3dstreet/pull/789/commits/69904017a9557f80e0881f8bb641fd655c50d28e

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,9 +17,9 @@ function Inspector() {
   this.history = require('./lib/history');
   this.isFirstOpen = true;
   this.modules = {};
-  this.on = Events.on;
-  this.emit = Events.emit;
-  this.off = Events.off;
+  this.on = Events.on.bind(Events);
+  this.emit = Events.emit.bind(Events);
+  this.off = Events.off.bind(Events);
   this.opened = false;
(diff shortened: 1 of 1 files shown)


vincentfretin:
Closed in favor of #755 
I changed my mind, the usage of `AFRAME.INSPECTOR.emit('entitycreated', el)` externally is wrong. The SceneGraph should be fixed to listen on child-attached/child-detached to update.

vincentfretin:
I reverted the above mentioned change in 3dstreet and did this instead https://github.com/3DStreet/3dstreet/pull/789/commits/5ac90b7647d7c31ade0438074fb2fb655b87e5ad
I'll do a similar change for aframe-inspector. This fixes showing in scene graph entities created by a component you add.

PR #755, opened 2024-08-28
[cleanup] Remove exposed on/emit/off that were not working properly

vincentfretin:
Remove exposed on/emit/off that were not working properly because of bad binding (see #754 for an explanation) so not used in the wild.
The `this.on` was there a long time ago, we introduced the `this.off` and `this.emit` in #719 but those weren't working.
@gregogalante you can still do your own build with #754 changes if you need that.

Instead of exposing the events publicly, let's keep it internal for now.
I want later to define documented commands that can be used instead (undo compatible).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,9 +17,6 @@ function Inspector() {
   this.history = require('./lib/history');
   this.isFirstOpen = true;
   this.modules = {};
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks

PR #757, opened 2024-08-30
[bugfix] Disable focusing an object with orthographic camera

vincentfretin:
Disable focusing an object with orthographic camera (used in top, left, right views) that put the camera at a wrong angle and in an unrecoverable state.


dmarcos:
Thanks

PR #759, opened 2024-08-31
Remove buggy ctrl+x to cut entity to paste it right away

vincentfretin:
Remove buggy ctrl+x to cut entity to paste it right away, because the entity is removed from the DOM, some components like position/rotation/scale reset to their default value.
If we really want to fix it, we should really keep an object representation of the deleted entity so that it could use createEntity(definition) when we press ctrl+v in this case instead of cloneEntity, but we currently don't support creating children on createEntity.
To be honest, I don't think it's worth to keep that feature, and I really don't understand what was the purpose of it in the first place.
We could keep it as way to remove an entity without confirmation, in this case we remove the `AFRAME.INSPECTOR.entityToCopy` assigment here and we reflect that purpose in the help modal. Thoughts?

Note that the ctrl+x,c,v shortcuts only worked on Windows since a long time, and were fixed for Linux and macOS only recently #729

### DIFF
--- a/src/components/modals/ModalHelp.js
+++ b/src/components/modals/ModalHelp.js
@@ -54,7 +54,7 @@ export default class ModalHelp extends React.Component {
         { key: ['ctrl | cmd', 'c'], description: 'Copy selected entity' },
         { key: ['ctrl | cmd', 'v'], description: 'Paste entity' },
         { key: ['h'], description: 'Show this help' },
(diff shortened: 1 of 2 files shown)


vincentfretin:
I found it weird we had "supr" here, in English keyboard it's Delete right. On my French keyboard, it's Suppr (two p) stands for Supprimer.

dmarcos:
Thank you!

PR #761, opened 2024-08-31
[bugfix] Update transform controls scale on camera change

### LINKED ISSUE #649: Transform controls really big

vincentfretin:
The scale of transform controls is now updated when you focus an element, zoom in, zoom out. This closes #649


dmarcos:
Thanks

PR #765, opened 2024-09-07
Update to GLTFExporter from three r168 which has better performance

vincentfretin:
This GLTFExporter change in r167 https://github.com/mrdoob/three.js/pull/28893 speed up one of my glb export from 7s to 3s. If you're using a canvas to create an image to download, you should really set { willReadFrequently: true } when getting the 2d context.


dmarcos:
Thanks!

PR #767, opened 2024-09-07
Don't export position/rotation/scale/visible if it's the empty value

### LINKED ISSUE #763: Don't export empty position/rotation/scale/visible
The position/rotation/scale/visible components are special, those don't store data, they just set things on object3D.
It's a bit silly to have visible="" exported (when you use copy entity to clipboard) if you toggled twice the visibility of the entity.
Same for position="" rotation="" scale="" it exports empty string when the value is the default value.

vincentfretin:
This fixes #763 

For example select the box, initially when you copy the entity to clipboard, you get
```html
<a-box id="box" position="0 0.91 0" height="2" color="#FFC65D"></a-box>
```

Then change the rotation to y 45 and mark it as not visible, if you copy the entity to clipboard:
```html
<a-box id="box" position="0 0.91 0" height="2" color="#FFC65D" visible="false" rotation="0 45 0"></a-box>
```

Then toggle back visible and reset rotation to 0.
Before the PR:
```html
<a-box id="box" position="0 0.91 0" height="2" color="#FFC65D" visible="" rotation=""></a-box>
```

After the PR:
```html
<a-box id="box" position="0 0.91 0" height="2" color="#FFC65D"></a-box>
```

### DIFF
--- a/src/lib/entity.js
+++ b/src/lib/entity.js
@@ -205,6 +205,17 @@ function optimizeComponents(copy, source) {
       var value = stringifyComponentValue(schema, optimalUpdate);
       setAttribute.call(copy, name, value);
     }
+
+    // Remove special components if they use the default value
+    if (
+      value === '' &&
+      (name === 'visible' ||
+        name === 'position' ||
+        name === 'rotation' ||
+        name === 'scale')
+    ) {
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks!

PR #772, opened 2024-09-17
Don't transpile to ES5

vincentfretin:
We used `@babel/preset-env` without browserlist config in `package.json`, so `@babel/preset-env` defaulted to ES5 and not browserlist `defaults` query that applies modern browsers. See doc
https://babeljs.io/docs/babel-preset-env
https://babeljs.io/docs/options#no-targets
but really nowadays we don't need to transpile to ES5, that was at a time we needed to support IE11.
We don't use any fancy javascript syntax here, only ES6/ES2015 that is supported by all supported browsers.

Bundle size reduced from 521K to 426K. Transpiling to ES5 included lots of polyfills we don't need.


vincentfretin:
I just had the idea to do this PR while watching https://www.youtube.com/watch?v=odhjF1obcUo

vincentfretin:
To use it:

```html
<a-scene inspector="url: https://cdn.jsdelivr.net/gh/aframevr/aframe-inspector@ddda46f4376cf696e623721d98e1c2eb3e799b73/dist/aframe-inspector.min.js">
```

vincentfretin:
We could also just remove the single test https://github.com/aframevr/aframe-inspector/blob/master/src/components/__tests__/Collapsible.test.js and jest, babel-jest, react-test-renderer dependencies instead of reintroducing @babel/preset-env in https://github.com/aframevr/aframe-inspector/pull/772/commits/f8f9035ae935c7d8416b0126fa6f326d50c8242f
@dmarcos thoughts?

dmarcos:
Yeah we don't need to transpile to ES5 I think. Is this ready to merge?

vincentfretin:
You can merge as is yes.

dmarcos:
Thanks so much!

PR #775, opened 2024-11-14
Be sure an entity doesn't use the same id as a mixin in the examples

vincentfretin:
`document.getElementById("blueBox")` returned the mixin and not the entity because both the mixin and the entity had the same id. I changed the id on the entity to not conflict. An id should be unique in an html page. @dmarcos


dmarcos:
Thank you!

PR #776, opened 2024-11-25
Reformat index.html

vincentfretin:
Reformat index.html for easier comparison with copy a-scene HTML to clipboard

@dmarcos


dmarcos:
Thanks!

PR #777, opened 2024-11-25
Remove setting data-aframe-inspector-original-camera attribute

vincentfretin:
Remove setting data-aframe-inspector-original-camera attribute that is not used anywhere.
This avoid this attribute to be exported when using copy entity html to clipboard.

@dmarcos


dmarcos:
Thanks!

PR #778, opened 2024-11-26
Update bitmovin url in 360video.html

vincentfretin:
@dmarcos same url as you did in aframe repo


dmarcos:
Thanks!

dmarcos:
wonder if we should find another 360 video that is under CC license that we can host in a-frame assets. this will bite us again in the future

PR #779, opened 2024-11-26
Reformat other examples, fix aframe master url, update to environment 1.4.0

vincentfretin:
@dmarcos It put back environment "fog: 0" in index.html I removed in #776 it's just that the copy to clipboard wrongly remove it because it's not aware of environment presets, only checking the schema default values.


dmarcos:
Thank you!

PR #780, opened 2024-11-26
In optimizeComponents be sure to also remove components that are defined in the primitive with empty object

vincentfretin:
In optimizeComponents used by the copy entity to clipboard feature, be sure to also remove components that are defined in the primitive with empty object, for example a-camera here https://github.com/aframevr/aframe/blob/a26768db5fe0f04a3b8fec3e73780b1855c3c75c/src/extras/primitives/primitives/a-camera.js#L5-L7

`defaultComponentsFromPrimitive` is defined by `this.defaultComponentsFromPrimitive = definition.defaultComponents || definition.defaultAttributes || {};` in aframe registerPrimitive. https://github.com/aframevr/aframe/blob/a26768db5fe0f04a3b8fec3e73780b1855c3c75c/src/extras/primitives/primitives.js#L37
`defaultComponents` is only on the definition, never on the html element, so checking for `source.defaultComponents` is always undefined from my understanding.
I modified `getInjectedValue` to check only `defaultComponentsFromPrimitive` and don't create unnessary object if it's undefined.

Also in `getMixedValue`, the line `source.mixinEls.reverse()` was reversing the mixinEls in place, so the order changed on each execution, not really what we want here, I replaced by `toReversed()`

`getImplicitValue` didn't return isInherited true in case of empty object in the primitive `defaultComponents`, so the component wasn't deleted as part of the `if (isInherited && doesNotNeedUpdate)` check.
https://github.com/aframevr/aframe-inspector/blob/ebb07fa84fe642a0f4112006c2c67a7b85c5cf85/src/lib/entity.js#L201-L202

I fixed it.

Before:

```html
<a-cursor raycaster cursor></a-cursor>
```

After:
```html
<a-cursor raycaster></a-cursor>
```

`raycaster` is set by the `cursor` component so that's why it still shows up, but I don't mind, it's good practice to define objects property in `raycaster` anyway.
  
Before:
```html
<a-camera camera look-controls wasd-controls>
```

After:
```html
<a-camera>
```

### DIFF
--- a/src/lib/entity.js
+++ b/src/lib/entity.js
@@ -284,6 +284,13 @@ function getImplicitValue(component, source) {
   function _multi() {
     var value;
 
+    // Set isInherited to true if the component is inherited from primitive defaultComponents and is empty object
+    var defaults = source.defaultComponentsFromPrimitive;
+    isInherited =
+      defaults &&
+      /* eslint-disable-next-line no-prototype-builtins */
+      defaults.hasOwnProperty(component.name) &&
+      Object.keys(defaults[component.name]).length === 0;
     Object.keys(component.schema).forEach(function (propertyName) {
       var propertyValue = getFromAttribute(component, propertyName, source);
       if (propertyValue === undefined) {
@@ -362,7 +369,7 @@ function getFromAttribute(component, propertyName, source) {
  */
 function getMixedValue(component, propertyName, source) {
   var value;
-  var reversedMixins = source.mixinEls.reverse();
+  var reversedMixins = source.mixinEls.toReversed();
   for (var i = 0; value === undefined && i < reversedMixins.length; i++) {
     var mixin = reversedMixins[i];
     /* eslint-disable-next-line no-prototype-builtins */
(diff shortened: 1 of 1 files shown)


vincentfretin:
@dmarcos that's a small part of my work on a better full html export of a-scene I'm currently doing.

vincentfretin:
in controllers.html with the mixin
```html
<a-mixin id="interactable" event-set__mouseenter="_event: mouseenter; material.opacity: 0.75" event-set__mouseleave="_event: mouseleave; material.opacity: 1" event-set__click="_event: click; scale: 1.1 1.1 1.1"></a-mixin>
```
The entity
```html
        <a-entity id="redBox" mixin="interactable" geometry="primitive: box" material="color: #F00" position="-4 1 0">
```
was exported to html with
```html
        <a-entity id="redBox" mixin="interactable" geometry="primitive: box" material="color: #F00" position="-4 1 0" event-set__mouseenter="[object Object]" event-set__mouseleave="[object Object]" event-set__click="[object Object]">
```
`event-set__mouseenter`, `event-set__mouseleave`, `event-set__click` shouldn't be there. Those didn't change from the mixin.
The code was checking `component.name` that is `event-set` instead of `component.attrName` (see https://github.com/aframevr/aframe/blob/a26768db5fe0f04a3b8fec3e73780b1855c3c75c/src/core/component.js#L68) that is the full attribute name `event-set__mouseenter`.

The serialization to `[object Object]` is a separate issue.

vincentfretin:
For the `event-set__mouseenter` serialization issue, see #781

dmarcos:
Thank you!

PR #781, opened 2024-11-26
Allow editing a custom type like event-set component schema

vincentfretin:
In the http://localhost:3333/examples/controllers.html example, the `event-set__click` component is wrongly serialized to "[object Object]"` in the right panel so we can't edit it.
To edit it properly we need the fix in this PR and in the event-set component there https://github.com/supermedium/superframe/pull/344

@dmarcos

### DIFF
--- a/src/components/components/PropertyRow.js
+++ b/src/components/components/PropertyRow.js
@@ -103,6 +103,14 @@ export default class PropertyRow extends React.Component {
         return <BooleanWidget {...widgetProps} />;
       }
       default: {
+        if (
+          props.schema.type === 'string' &&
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks!

vincentfretin:
Please don't forget to also look at https://github.com/supermedium/superframe/pull/344 and make a new release.

dmarcos:
I published a new version of the event-set component. Are there any other components that need a new release? Thanks

vincentfretin:
I think that was the only component that needed a release. Thanks!

PR #782, opened 2024-11-26
Add support for comparing property of type selector

vincentfretin:
Add support for comparing property of type selector, this fixes the error 'RangeError: Maximum call stack size exceeded' that you get on `Object.keys(playerEntity)` when using copy entity HTML to clipboard with a component having a selector property like movement-controls's camera property.


dmarcos:
Thanks!

PR #783, opened 2024-12-12
Emit entityselect even if entity is null so the React UI updates properly

vincentfretin:
When you press escape it does `AFRAME.INSPECTOR.selectEntity(null)` but the entityselect event wasn't emitted so the UI didn't update properly, and the right panel stayed open instead of closing.
@dmarcos


dmarcos:
Thank you!

PR #785, opened 2025-01-13
Update ghpages command to use gh-pages instead of unmaintained ghpages

vincentfretin:
https://github.com/cvan/ghpages is not maintained (6 vulnerabilities reported) and is just a wrapper to https://github.com/tschaub/gh-pages
Update the command to use gh-pages directly.
I tested the `npm run deploy` command, that works properly like before.

Before the change:
```
12 vulnerabilities (1 low, 3 moderate, 6 high, 2 critical)
```

After the change:
```
6 vulnerabilities (1 low, 3 moderate, 2 high)
added 11 packages, removed 54 packages, changed 5 packages, and audited 1201 packages in 7s
```
and then after doing `npm audit fix` to fix the remaining vulnerabilities:
```
found 0 vulnerabilities
```


dmarcos:
thanks

PR #786, opened 2025-01-14
Textures modal: remove dead code loadedTextures

vincentfretin:
That code is never used, loadedTextures always stays an empty array.


dmarcos:
Thank you!

PR #789, opened 2025-01-14
Increase width of camera select a bit so that Perspective is not cut

vincentfretin:
Before:
image

After:
image


dmarcos:
Thanks!

PR #790, opened 2025-01-14
Remove unused css for classes help-link dropbtn dropdown dropdown-content

vincentfretin:
Those classes are used nowhere.


dmarcos:
Thanks!

PR #792, opened 2025-01-14
Add sponsor button

vincentfretin:
Add a sponsor button next to the Back to Scene button that opens a modal like this:
image
button on hover:
image
modal:
image


dmarcos:
good luck!

PR #797, opened 2025-01-27
Fix copy component to clipboard

### LINKED ISSUE #672: Copy component values to clipboard wrong
Copy component values (values that are different from defaults) to clipboard is not working correctly for `geometry` and `gltf-model`, I get just `geometry=""`, `gltf-model=""`. For `material` component, I get all props with `[object Object]`.
I didn't see an issue with my custom components.
The copy entity to clipboard implementation is correct for all three components.

I won't investigate this right away.

vincentfretin:
Fix copy component to clipboard.
This closes #672

Click on the yellow sphere and then click on copy to clipboard for material, we properly get now:
`material="color: #ff0; roughness: 1"`

Same as copy entity to clipboard:
```html
<a-entity id="yellowSphere" geometry="primitive: sphere" material="color: #ff0; roughness: 1" position="-2 2 -2" models-array="models: one, two"></a-entity>
```

### DIFF
--- a/src/lib/entity.js
+++ b/src/lib/entity.js
@@ -500,35 +500,14 @@ function getUniqueId(baseId) {
 }
 
 export function getComponentClipboardRepresentation(entity, componentName) {
-  /**
-   * Get the list of modified properties
-   * @param  {Element} entity        Entity where the component belongs
-   * @param  {string} componentName Component name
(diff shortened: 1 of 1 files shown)


dmarcos:
Super thanks

PR #798, opened 2025-01-27
Show properties of a component only if allowed by schema condition

### LINKED ISSUE #762: Don't show irrelevant properties of a component based on the schema constraint
Changing groundColor of pointLight gives an error at line
at 
https://github.com/aframevr/aframe/blob/80dd9507dc5ff615e8ca0d38df838cbb87eca052/src/components/light.js#L84
on
```
light.groundColor.set(value);
```
with light.groundColor undefined because that exists only for hemisphere.

The right panel shouldn't show the groundColor property, it should take into account the constraint defined in schema to show the properties
https://github.com/aframevr/aframe/blob/80dd9507dc5ff615e8ca0d38df838cbb87eca052/src/components/light.js#L22
```
    groundColor: {type: 'color', if: {type: ['hemisphere']}},
```

vincentfretin:
Show properties of a component only if allowed by schema condition.
For example for light component:
- only show castShadow checkbox if type is point, spot or directional
- show additional shadow properties when castShadow checked
- show groundColor only if type is hemisphere

This closes #762

### DIFF
--- a/src/components/components/Component.js
+++ b/src/components/components/Component.js
@@ -94,6 +94,30 @@ export default class Component extends React.Component {
 
     return Object.keys(componentData.schema)
       .sort()
+      .filter((propertyName) => {
+        if (!componentData.schema[propertyName].if) {
+          return true;
+        }
+        let showProperty = true;
+        for (const [conditionKey, conditionValue] of Object.entries(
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks

PR #799, opened 2025-01-27
Pause the cloned entity in the loaded listener

### LINKED ISSUE #690: Cloning an entity with animation animates, it shouldn't
If you're cloning an entity with "d" key and the entity has an animation, it animates, it shouldn't. The animations should be paused while in editor mode.

vincentfretin:
Pause the cloned entity in the loaded listener so the animation on red cube doesn't play.
This closes #690


dmarcos:
Thanks!

PR #800, opened 2025-02-04
Be sure to use a valid id for multiple component, remove dead code optionRenderer

vincentfretin:
For the add component select, be sure to use a valid id for multiple component.
Also remove dead code optionRenderer that doesn't exist in the react-select version we're using and that custom option render was to add an icon if a component was loaded, they all are! I think it was a thing of the past with angle/aframe registry.


dmarcos:
Thanks!

PR #805, opened 2025-02-21
Fix console warning about using SelectWidget with a number as value

vincentfretin:
Fix console warning about using SelectWidget with a number as value, that works fine.


dmarcos:
Thank you!

PR #806, opened 2025-04-05
Fix error when material component uses a canvas texture and add an example

vincentfretin:
When selecting an entity that has the material component referencing a canvas texture, we get the error:
```js
TypeError: t.match is not a function
    at Module.fE (src-loader.js:114:23)
    at Jr.setValue (TextureWidget.js:149:36)
    at Jr.componentDidMount (TextureWidget.js:88:10)
```
The code goes in
https://github.com/aframevr/aframe-inspector/blob/ea5c26931e8b9f564166020540113580dee9d19b/src/components/widgets/TextureWidget.js#L149
expecting to extract an url from a string value `url(someurl)` but here value is a canvas element.

The fix here is to check for canvas element similar to img or video. The url is only shown in the tooltip, for canvas we don't have an url.

I added an example with a smiley draw on a canvas:
Capture d’écran du 2025-04-05 13-25-44

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -12,11 +12,60 @@
         models: {type: 'array', oneOf: ['one', 'two', 'three', 'four']}
       }
     })
+
+    AFRAME.registerComponent('draw-smiley', {
+      schema: {
+        canvas: {type: 'selector'}
+      },
+      init() {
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks

PR #819, opened 2025-08-01
Fix clicking on boolean property label to toggle the checkbox

vincentfretin:
Remove unused widgets propTypes, fix clicking on boolean label and add id to other widgets, simplify a bit the logic in PropertyRow that overrides some type and value


dmarcos:
thanks

PR #820, opened 2025-08-22
Fix wrong camera position calculation when focusing an element

vincentfretin:
The calculation I did in #760 was not quite right.
The center was assumed to be in local coordinates but it is actually already in world coordinates, so localToWorld increased the y value so the camera was much too high. The issue is not much visible in a scene where all models are on the ground, it's much more visible on a model in the air. Even for models on the ground, that fixes the glitch when you start rotating after focusing an element.

Before
focus_before

After:
focus_after

### DIFF
--- a/src/lib/EditorControls.js
+++ b/src/lib/EditorControls.js
@@ -58,27 +58,27 @@ THREE.EditorControls = function (_object, domElement) {
       return;
     }
     var distance;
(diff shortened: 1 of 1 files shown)


vincentfretin:
@dmarcos this is ready to be merged

dmarcos:
thank you!

PR #821, opened 2025-09-30
Properly close canvas element in index.html

vincentfretin:
Properly close canvas element in index.html, we're in html, not in jsx.
@dmarcos


dmarcos:
Thank you!

PR #822, opened 2025-10-17
Allow dash in component instance id like bind__troika-text

vincentfretin:
I wanted to use https://github.com/supermedium/superframe/tree/master/components/state
so I entered `troika-text` when adding the `bind` component, but it wrongly created `bind__troikatext` instead of `bind__troika-text` and so the bind component didn't work.
A dash in component instance id is allowed.
This PR fixes the issue @dmarcos


dmarcos:
Thank you!

PR #824, opened 2025-11-09
Be sure to not register again asset with same url, create a-assets if missing

vincentfretin:
Steps to reproduce:
- select yellow cube
- go to the material's src property and click the button to open the textures modal
- right click on crate image and copy url
- click LOAD TEXTURE
- paste the url and press enter (or you can also click on the crate image)
- it currently shows the "crate" name and pressing LOAD THIS TEXTURE will add a second time that asset with a different name. That's wrong.

What it should do is showing the existing asset id "crateImg" and not "crate" to not register the same asset a second time.
I now show the message "Texture already loaded" and the "LOAD THIS TEXTURE" button stays disabled in this case.
I also now handle the case "Name already taken by another asset or entity".

The code 

```js
if (image.type !== 'asset') {
  var assetId = insertOrGetImageAsset(image.src);
  value = '#' + assetId;
}
```
that is executed after selecting a texture that closes the modal probably took care of reusing the same asset in the past but it was a dead branch in the current code, `image.type` is always "asset" here because `props.onClose(value)` is called from `selectTexture` that is used only on registered assets.
I removed that code and the `insertOrGetImageAsset` function and reintroduced a check in the load texture step to check for existing asset with the same url.
I removed the duplicate `insertNewAsset/getFilename` functions and moved `getFilename`, `isValidId`, `getValidId`, `getUrlFromId` and added new `getIdFromUrl` to `assetsUtils.js` to be reused with other type of assets and not only img in the future (like selecting an audio for the sound component).
The `a-assets` element is now created if it doesn't exist, it gave an error previously.
The input field and preview is now cleared if you click on the BACK button.

### DIFF
--- a/src/components/modals/ModalTextures.js
+++ b/src/components/modals/ModalTextures.js
@@ -347,11 +347,11 @@ export default class ModalTextures extends React.Component {
               />
               {preview.type !== 'asset' && assetIdTaken && (
                 <div className="iderror">
-                  id already taken by another asset or entity
+                  Name already taken by another asset or entity
                 </div>
               )}
               {this.state.preview.name.length > 0 && !validId && (
-                <div className="iderror">id is not valid</div>
+                <div className="iderror">Name is not valid</div>
               )}
               {preview.type === 'asset' && (
                 <div className="iderror">Texture already loaded</div>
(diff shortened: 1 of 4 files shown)


vincentfretin:
Capture d’écran du 2025-11-10 14-00-30

Capture d’écran du 2025-11-10 14-00-59

Capture d’écran du 2025-11-10 13-51-31

PR #830, opened 2025-12-06
Fix some vertical alignment issues, Safari style issues, select and checkbox now uses the primary color

vincentfretin:
- Remove unused rules
- Set border-radius 0 on input to fix issue on Safari where input id appeared rounded and fix the color picker to be rectangle instead of ovale on Safari
- Use custom checkbox style so it appears the same on Chrome, Firefox, Safari with the correct primary color
- Specify a .select__option--is-selected rule to use the correct primary color in camera select
- Fix vertical alignment issues with entityPrint and icons rendered in scene graph and sidebar header, fix search input to make the scene graph horizontal scroll disappear
- Increase font-weight for .propertyRowDefined to be more visible

on Chrome
Before
before

After
after

### DIFF
--- a/src/style/components.styl
+++ b/src/style/components.styl
@@ -39,25 +39,6 @@
   font-size 13px
   &:hover
     background $bglighter
-/*
-.collapsible
-  &.collapsed
-    background-color $grayalt
-    .static,
-    .componentHeaderActions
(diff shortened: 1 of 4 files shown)


dmarcos:
thanks!

PR #837, opened 2025-12-20
Fix modal background and search icon

vincentfretin:
While doing the stylelint integration #836 I didn't pay attention that stylus doesn't support the new css rgb syntax `rgb(0 0 0 / 50%)` so the modal background was actually just black, we still need to use the stylus rgba function, so I reverted that change and ignored some rules in stylelint config.
I also removed the non existing animatetop animation, fixed the search icon position, and increased the font weight of the modal title.

texturemodal

### DIFF
--- a/src/style/index.styl
+++ b/src/style/index.styl
@@ -169,7 +169,7 @@ body.aframe-inspector-opened
     z-index 999999999
 
     svg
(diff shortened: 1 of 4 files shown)


dmarcos:
thanks

PR #854, opened 2026-04-18
Fix editor focus for scaled targets

vincentfretin:
EditorControls.focus used target.localToWorld to place the camera, which multiplied the world-space bbox-derived offset by the target's world scale. Focusing a model scaled by N ended up N times too far. Decompose matrixWorld instead and rotate the offset by the target's world rotation only.


dmarcos:
thank you!

PR #855, opened 2026-04-23
Support editing properties of type array without oneOf, selector and selectorAll

vincentfretin:
Add support for editing a property of type array without oneOf like animation startEvents, selector and selectorAll types.

I did the same in aframe-editor https://github.com/c-frame/aframe-editor/pull/69 already released in [1.7.25](https://github.com/c-frame/aframe-editor/releases/tag/1.7.25)


dmarcos:
thanks

PR #856, opened 2026-04-25
Fix crash in selectionBox when gltf-model loads after entity is detached

vincentfretin:
Very rare case but could happen when you undo/redo a reparent action very quickly in aframe-editor. Here in the inspector that's probably even harder to reproduce, maybe copy an entity having gltf-model, paste and delete enter right away but with the confirmation modal it leave enough time to not produce the error.


dmarcos:
thank you!

PR #859, opened 2026-07-04
Material assets support: property widget, materials modal, asset deletion

vincentfretin:
## Description

Companion to aframevr/aframe#5846 (materials as assets and property type).

### Material property widget

The new `material` property type references an `<a-material>` asset by selector (`#myMaterial`) or an inline `material(...)` definition, and its parsed value is a `THREE.Material`. Without these changes, the property fell through to the generic `InputWidget`, which parses and commits on **every keystroke** — typing a partial selector like `#crate` immediately committed a null material and detached the entity from its shared material while typing.

The property is now rendered by a dedicated `MaterialWidget`: the raw reference string committed on blur (like `selector`/`selectorAll`), plus a swatch previewing the referenced material's color/texture.

### Material Assets modal (picker + live editor)

Clicking the swatch opens a Material Assets modal that:

- lists every `<a-material>` in the scene with a color/texture swatch,
- applies the selected asset to the property (USE SELECTED button or double-click),
- edits the selected material's properties with schema-driven widgets (base material + shader schema, `shader` shown read-only since it cannot change after creation) — edits apply live to every entity sharing the material; map properties use the texture picker, stacked on top of the materials modal (`Modal` is now stacking-aware: only the topmost open modal reacts to ESC/outside clicks),
- creates new `<a-material>` assets under `<a-assets>` (NEW MATERIAL button),
- deletes the selected material asset (DELETE button, disabled for inline materials; the confirmation mentions how many entities use the material).

The textures modal also gets a trash button on each asset image to remove it from `<a-assets>`.

The example scene includes two shared `<a-material>` assets, an inline definition, and entities using them (requires an A-Frame build with aframevr/aframe#5846; with older builds these entities render with a default material and a warning).

Verified against an aframe build of the PR branch: picking switches the entity's shared material (textures included), editing roughness/color in the modal updates all entities sharing the material live, the reset button returns the entity to its own component-managed material, and typing partial selectors no longer detaches the material.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -77,6 +77,8 @@
         <a-mixin id="purple" material="color: #93648D"></a-mixin>
         <a-mixin id="short" scale="1 0.5 1"></a-mixin>
         <a-mixin id="yellow" material="color: #FFC65D"></a-mixin>
+        <a-material id="gold" color="#FFC65D" metalness="0.9" roughness="0.2"></a-material>
+        <a-material id="crateMat" src="#crateImg" roughness="0.9"></a-material>
         <img id="crateImg" src="https://aframe.io/sample-assets/assets/images/wood/crate.gif" crossorigin="true">
         <img id="floorImg" src="https://aframe.io/sample-assets/assets/images/terrain/grasslight-big.jpg" crossorigin="true">
       </a-assets>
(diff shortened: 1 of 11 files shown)


vincentfretin:
Capture d’écran du 2026-07-04 11-10-53

Capture d’écran du 2026-07-04 11-11-25

